PR #17, opened 2019-03-08
Add perm_ternary

danebou:


strip_other_fns.py
```diff
@@ -1,28 +1,57 @@
...
+        f.write(stripped)
 
 if __name__ == "__main__":
     filename = sys.argv[1]
     fn_name = sys.argv[2]
-    strip_other_fns(filename, fn_name)
+    
+    source = Path(filename).read_text()
+    strip_other_fns_and_write(source, fn_name)
     pass
```
simonlindholm: Why do we need to rewrite this based on source instead of AST?

preprocess.py
```diff
@@ -0,0 +1,7 @@
+from pycparser import preprocess_file
+import re
+
+def preprocess(file, cpp_args=''):       
+    preprocessed = preprocess_file(file, cpp_args=cpp_args)
+    preprocessed = re.sub(r'#.*\n', '', preprocessed)
```
simonlindholm: instead of doing this, we could pass `-P` as cpp_args (`-nostdinc` also inhibits some line markers). With that in mind, does this deserve a whole new file and function? Do we have plans to grow it?

I guess if we start auto-injecting macros, e.g. if we start basing more stuff on AST traversals, then that would warrant a function... but we don't want such macros from where we call this function from test code.

strip_other_fns.py
danebou: So we can pull off this: `PERM_TERNARY(test = ,cond,1,2)`

simonlindholm: Oh. Right, that makes sense. An alternative would be to first convert all arguments to our special macros into strings, then do AST magic, then unstringify them. Not sure if that's better though.

preprocess.py
danebou: I'm new to python code structure, but at least with other other languages I go with the philosophy that if it can be though of as a object, it can be put in its own file. (Information hiding and reuse benefits).
I'll make the `-P` change

strip_other_fns.py
danebou: It didn't compile as an AST

preprocess.py
simonlindholm: Agreed in general, but yeah, it's a bit silly for a single one-line function that we don't expect to grow any further.

PR #18, opened 2019-03-08
Fix infinite loop in decl randomizer

danebou:
This occurs if the only declaration is the function declaration


randomizer.py
```diff
@@ -613,14 +613,17 @@ def perm_randomize_type(fn: ca.FuncDef, ast: ca.FileAST) -> bool:
...
     Visitor().visit(fn)
-    while True:
-        decl = random.choice(decls)
-        if isinstance(decl.type, ca.TypeDecl):
-            decl.type = randomize_type(decl.type, typemap)
-            set_decl_name(decl)
-            break
+
+    if len(decls) == 0:
+        return True
```
simonlindholm: make this `return False`

simonlindholm:
Merged with that fix.

PR #23, opened 2019-03-14
Dependency inject random

danebou:
Use dependency injection for the random object.

This is in preparation for multi-threading so each thread can get its own random class and concurrency issues.

### DIFF
--- a/main.py
+++ b/main.py
@@ -3,7 +3,7 @@ import argparse
 import difflib
 import functools
 import os
-import random
+from random import Random
 import re
 import sys
 import time
@@ -35,7 +35,7 @@ def find_fns(source: str) -> List[str]:
     return [fn for fn in fns if not fn.startswith('PERM')]
 
 class Permuter:
-    def __init__(self, dir: str, compiler: Compiler, scorer: Scorer, source_file: str, source: str):
+    def __init__(self, dir: str, compiler: Compiler, scorer: Scorer, source_file: str, source: str, seed: int):
         self.dir = dir
         self.compiler = compiler
         self.scorer = scorer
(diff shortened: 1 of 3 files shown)


danebou:
> Need to fix seeding first

should be good now!

main.py
```diff
@@ -117,8 +119,8 @@ def write_candidate(perm: Permuter, source: str) -> None:
 
 def main(options: Options) -> int:
     last_time = time.time()
-    if options.seed is not None:
-        random.seed(options.seed)
+    if options.seed is None:
+        options.seed = os.urandom(8)
```
simonlindholm: this is bugging me for two reasons:
- options.seed is typed as an Optional[int], but here we assign bytes
- further down we check `if iteration == 0 and options.seed is not None`; now `options.seed is not None` cannot happen

Maybe we can type `options.seed` as int to resolve the second point and make the caller do the random seed defaulting work, if the goal is to have wrapped_main not depend on global randomness state? Otherwise you can move this seeding into that function.

danebou: Yeah, this should be better now.

danebou: Give me one sec, need to fix the initial seed setter

danebou: Better now. Each permuter keeps track of its own current seed for each iteration. When initialized, when initialized every permuter starts with the same seed.

danebou: (I also tested it this time, and get the same score updates)

main.py
```diff
@@ -140,6 +149,12 @@ def heartbeat() -> None:
 def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
     print("Loading...")
 
+
+    if options.seed != None:
+        start_seed = options.seed
+    else:
+        start_seed = randrange(sys.maxsize) # Random seed from default random
```
simonlindholm: import random, random.randrange

main.py
```diff
@@ -140,6 +149,12 @@ def heartbeat() -> None:
 def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
     print("Loading...")
 
+
+    if options.seed != None:
```
simonlindholm: `is not None` to make mypy happy (and it's more Pythonic, I believe)

main.py
```diff
@@ -74,17 +79,23 @@ def score_base(self) -> Tuple[str, int, str]:
 
         return (base_source, *self.scorer.score(start_o))
 
+    def _update_seed(self) -> None:
+        rand_seed = self.random.randrange(10**18) + self.iteration
+        self.random.seed(rand_seed)
```
simonlindholm: No point in doing this reseeding if we don't save the new seed (which we shouldn't since it's somewhat broken, as discussed).

PR #25, opened 2019-03-20
Parallelization

danebou:
Still WIP. Have to get working, cleanup and properly merge everything.


danebou:
Still need to split this up, but then should be good

danebou:
Should be good now!

main.py
```diff
@@ -30,15 +37,106 @@ class Options:
...
+        return p
+
+    def reset_ast(self) -> None:
+        self.ast.ext[self.fn_index] = copy.deepcopy(self.orig_fn)
+        self._cache_source = None
+
+    def randomize_ast(self, randomizer: Randomizer) -> None:
+        if randomizer.random.uniform(0, 1) >= RANDOMIZER_KEEP_PROB:
+            self.reset_ast()
+        randomizer.random.seed(self.seed)
```
simonlindholm: this if statement implies that instances of this class live for while... but you're resetting to the random seed repeatedly. that sounds like a bug?

main.py
```diff
@@ -30,15 +37,106 @@ class Options:
...
+        self._remove_o_file()
+
+class Profiler(object):
+    class StatType(Enum):
+        perm = 1
+        compile = 2
+        score = 3
+
+    def __init__(self):
+        self.time_stats = dict([(x, 0.0) for x in Profiler.StatType])
```
simonlindholm: `{x: 0.0 for x in Profiler.StatType}`

main.py
```diff
@@ -30,15 +37,106 @@ class Options:
...
+        if self.o_file is not None:
+            try:
+                os.remove(self.o_file)
+            except:
+                pass
+
+    def __del__(self) -> None:
+        self._remove_o_file()
+
+class Profiler(object):
```
simonlindholm: you don't need `(object)` I believe

(nice refactoring here btw)

main.py
simonlindholm: or `defaultdict(float)` even, if we want to support flexible stat types (up to you)

main.py
```diff
@@ -122,7 +230,81 @@ def write_candidate(perm: Permuter, source: str) -> None:
         except FileExistsError:
             pass
     print(f"wrote to {fname}")
-
+@attr.s
+class EvalContext():
```
simonlindholm: not used

main.py
```diff
@@ -51,62 +149,72 @@ def __init__(self, dir: str, compiler: Compiler, scorer: Scorer, source_file: st
...
-        return self.compiler.compile(self.get_source())
+        t3 = time.time()
 
-    def score(self, cand_o: str) -> Tuple[int, str]:
-        return self.scorer.score(cand_o)
+        profiler: Profiler = Profiler()
+        profiler.add_stat(Profiler.StatType.perm, t1 - t0)
+        profiler.add_stat(Profiler.StatType.compile, t2 - t1)
+        profiler.add_stat(Profiler.StatType.score, t3 - t2)
+        timings = profiler.get_str_stats()
```
simonlindholm: creating a new profiler every time makes the results jitter, ideally we should be reusing a single one instead

main.py
```diff
@@ -175,90 +370,56 @@ def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
...
+                try:
+                    cand, profiler = permuter.eval_candidate(seed)
+                except Exception as e:
+                    cand = None
+                    profiler = None
+                    exception = e
+                post_score(permuter, cand, exception, profiler)
+        else:
+            # Run multi-threaded
+            def done(f: Future, permuter: Permuter) -> None:
```
simonlindholm: not used

main.py
```diff
@@ -141,19 +323,32 @@ def heartbeat() -> None:
...
         # due to GC. But since we're sane people and don't use finalizers for
         # cleanup, we can just skip GC and exit immediately.
-        os._exit(0)
+        #os._exit(0)
+
+iteration = 0
+errors = 0
+overall_profiler = Profiler()
+high_scores = {}
+permuters: List[Permuter] = []
```
simonlindholm: global state :( can we avoid it?

danebou: Someone smarter probably can, but I ran into a lot of issues passing context to a new thread as opposed to it being forked :(

main.py
danebou: I'd agree, but then we'd have to do some nasty cross-process communication. The profiling has to happen separately on each new thread, and somehow needs to get joined back up. Reusing the profiler class seems the best way?

main.py
```diff
@@ -121,8 +162,69 @@ def write_candidate(perm: Permuter, source: str) -> None:
...
+        source = cand.get_source()
+        write_candidate(permuter, source)
+    print("\b"*10 + " "*10 + "\r" + status_line, end='', flush=True)
+    
+    # Cleanup cand
+    if cand:
+        with cand:
+            pass
 
+def gen_all_seeds(permuters: List[Permuter], heartbeat: Callable[[], None]) -> Iterable[Tuple[int, Permuter]]:
```
simonlindholm: remove the heartbeat argument

main.py
```diff
@@ -175,90 +282,72 @@ def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
...
+        print(f"[{permuter.unique_name}] base score = {permuter.base.score_value}")
+
+    context.high_scores = dict([(p.base.score_value, p) for p in context.permuters])
+
+    perm_seed_iter = gen_all_seeds(context.permuters, heartbeat)
+    if options.threads == 1:
+        for permuter_index, seed in perm_seed_iter:
+            heartbeat()
+            permuter = context.permuters[permuter_index]
+            # Run single threaded
```
simonlindholm: slightly misplaced comment (and somewhat of a pointless comment)

main.py
```diff
@@ -175,90 +282,72 @@ def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
...
+                except e:
+                    result = (None, None, permuter_index, e)
+                output.put(result)
+
+        # Begin workers
+        for i in range(options.threads):
+            Process(target=worker, args=(task_queue, results_queue)).start()
+        
+        # Run batch work
+        for bat in iter_util.batch(iter(perm_seed_iter), options.threads * 100):
```
simonlindholm: any reason to do batching with the persistent worker setup? seems like you could just as well put #worker * 2 or so tasks in the queue initially, and every time an item is finished, put in the next. Something like:

```
active_items = 0
q = Queue()
def more_work():
  nonlocal active_items, q
  work = next(iter, None)
  if work is not None:
    q.put(work)
    active_items += 1

for i in range(#workers * 2):
  more_work()

while active_items > 0:
  heartbeat()
  # (wait for one)
  active_items -= 1
  more_work()
```

main.py
```diff
@@ -175,90 +282,72 @@ def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
...
+                profiler = None
+                exception = e
+            post_score(context, permuter, cand, exception, profiler)
+    else: # Run multi-threaded
+        # Create queues
+        task_queue = Queue()
+        results_queue = Queue()
+
+        def worker(input, output):
+            global context
```
simonlindholm: does it work to make it `nonlocal`? still ugly, but perhaps less so...

main.py
```diff
@@ -175,90 +282,72 @@ def wrapped_main(options: Options, heartbeat: Callable[[], None]) -> int:
...
+                cand = None
+                profiler = None
+                exception = e
+            post_score(context, permuter, cand, exception, profiler)
+    else: # Run multi-threaded
+        # Create queues
+        task_queue = Queue()
+        results_queue = Queue()
+
+        def worker(input, output):
```
simonlindholm: type annotations would be useful

main.py
```diff
@@ -9,15 +9,25 @@
 import sys
 import time
 import traceback
+from multiprocessing import Queue, Process
```
simonlindholm: I would do `import multiprocessing` and use `multiprocessing.Queue` etc. further down; I think that makes it clearer what kind of queue is being used

main.py
```diff
@@ -121,8 +162,69 @@ def write_candidate(perm: Permuter, source: str) -> None:
             break
         except FileExistsError:
             pass
-    print(f"wrote to {fname}")
+    print(f"wrote to {fname}")   
+    
+def post_score(context: EvalContext, permuter: Permuter, cand: Candidate, exception: BaseException, profiler: Profiler) -> None:
+    if exception != None:
+        print(f"[{permuter.unique_name}] internal permuter failure.")
+        traceback.print_exc()
```
simonlindholm: have you tested that you can send exceptions across processes? otherwise, we could send traceback strings instead

main.py
```diff
@@ -121,8 +162,69 @@ def write_candidate(perm: Permuter, source: str) -> None:
...
+        print(f"To reproduce the failure, rerun with: --seed {permuter.seed}")
+        exit(1)
+
+    if options.print_diffs:
+        permuter.print_diff(cand)
+        yn = input("Press any key to continue...")
```
simonlindholm: fair enough, ctrl+c works. remove the `yn` var though

main.py
```diff
@@ -121,8 +162,69 @@ def write_candidate(perm: Permuter, source: str) -> None:
...
+
+    context.iteration += 1
+    if cand.score_value is None:
+        context.errors += 1
+    disp_score = 'inf' if cand.score_value == permuter.scorer.PENALTY_INF else cand.score_value
+    timings = ''
+    if options.show_timings:
+        assert(profiler)
```
simonlindholm: `profiler` is not Optional, so this assert doesn't seem meaningful assuming the program type-checks

(also, remove parentheses, assert is a keyword)

main.py
```diff
@@ -121,8 +162,69 @@ def write_candidate(perm: Permuter, source: str) -> None:
...
+        for stattype in profiler.time_stats:
+            context.overall_profiler.add_stat(stattype, profiler.time_stats[stattype])
+        timings = context.overall_profiler.get_str_stats()
```
simonlindholm: `'\t' + context.overall_profiler.get_str_stats()`, so we don't have to end the line with a trailing tab on the next line

main.py
```diff
@@ -121,8 +162,69 @@ def write_candidate(perm: Permuter, source: str) -> None:
...
+    status_line = f"iteration {context.iteration}, {context.errors} errors, score = {disp_score}\t{timings}"
+
+    if cand.score_value and cand.score_value <= permuter.base.score_value and cand.score_hash and cand.score_hash not in permuter.hashes and cand.score_value < 20000:
```
simonlindholm: 20000? I can agree that large scores are often spammy, but we should really use a low base score to get rid of them, and perhaps require score <= some constant * base score for large base scores (perhaps with a command-line configurable constant?)

main.py
```diff
@@ -143,17 +245,22 @@ def heartbeat() -> None:
         # cleanup, we can just skip GC and exit immediately.
         os._exit(0)
 
+# This has to be a global since we need to access the permuters. 
+# multiprocessing fork()'s the main thread and it works out that the permuters
+# are unmodified because we are simply passing in pre-calculated seeds from the
+# main thread. In theory, we should be able to pass in the context as an argument
+# but haven't been able to get pass various exceptions (probably related to pickling).
```
simonlindholm: sounds odd, given that Candidates apparently can be sent through task queues... what exception do you get?

main.py
danebou: whoops. That was debug code!

simonlindholm:
I'm currently working on cleaning this up a bit, hopefully we can get it merged soon.

simonlindholm:
Hah, the unit tests just caught a terrible bug:
```
    if score_value and score_value <= permuter.base_score and score_hash not in permuter.hashes:
        ...
        write_candidate(permuter, source)
```
matches with score 0 were never written out!

danebou:
Been a while but thanks for the massive cleanup while I'm on hiatus!

PR #28, opened 2019-04-12
Line swap perm

danebou:


perm/perm.py
```diff
@@ -160,3 +161,16 @@ def evaluate(self, seed: int, state: EvalState) -> str:
             return f'{cond}'
         else:
             return f'({cond}) != 0'
+
+class LineSwapPerm(Perm):
+    def __init__(self, lines: List[Perm]) -> None:
+        super().__init__()
+        self.children = lines
+        self.line_permutations = [i for i in itertools.permutations(range(len(lines)))]
+        self.perm_count = len(self.line_permutations)
```
simonlindholm: hm, I missed this for CondNezPerm as well, but this needs to have `* count_all(self.children)`

PR #29, opened 2019-04-12
Change split character from , to |

danebou:
This is to better support the PERM_LINESWAP, but more generally other C expressions which use `,`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The permuter expects as input one or more directory containing:
 
 See USAGE.md for more details on how to set these up.
 
-The .c file may be modified with any of the following macros which affect manual permutation. Arguments to the macros are split by `|`:
+The .c file may be modified with any of the following macros which affect manual permutation. Arguments to the macros are split by `|` (when not nested within parenthesis):
 
 - `PERM_GENERAL(a|b|...)` expands to any of `a`, `b`, ...
 - `PERM_TYPECAST(a|b|...)` expands to any of `(a)`, `(b)`, ... (empty argument for no cast at all)
--- a/README.md
+++ b/README.md
@@ -17,19 +17,19 @@ The permuter expects as input one or more directory containing:
 
 See USAGE.md for more details on how to set these up.
 
-The .c file may be modified with any of the following macros which affect manual permutation:
+The .c file may be modified with any of the following macros which affect manual permutation. Arguments to the macros are split by `|`:
 
-- `PERM_GENERAL(a, b, ...)` expands to any of `a`, `b`, ...
-- `PERM_TYPECAST(a, b, ...)` expands to any of `(a)`, `(b)`, ... (empty argument for no cast at all)
-- `PERM_TERNARY(prefix, a, b, c)` expands to either `prefix a ? b : c` or `if (a) prefix b; else prefix c;`.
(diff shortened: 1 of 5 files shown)


test/test_type.c
```diff
@@ -1,29 +1,29 @@
 #ifdef ACTUAL
-int test_type1(int a, int b) {
+int test_type1(int a| int b) {
```
simonlindholm: over-eager search-replace

README.md
```diff
@@ -17,19 +17,19 @@ The permuter expects as input one or more directory containing:
 
 See USAGE.md for more details on how to set these up.
 
-The .c file may be modified with any of the following macros which affect manual permutation:
+The .c file may be modified with any of the following macros which affect manual permutation. Arguments to the macros are split by `|`:
```
simonlindholm: should say something about `|` nested within parentheses not being split at

test/test_type.c
danebou: Whoops

danebou:
> Looks a bit ugly, but I suppose it's reasonable... I wonder if it could conflict with bitor or logical or. We could possibly work around the latter by saying that the `|` must not be adjacent to another `|`... or we can accept that parentheses are enough to prevent `|` from being parsed as separators.

Good point. You mean like where `||` translates to the literal character `|` while just `|` gets interpreted as the argument splitter? Or some type of escape sequence?

simonlindholm:
I was thinking that `||` maybe never occurs as part of a PERM calls, so could be special-case it to mean `||` (or `|` as an escape sequence). But on second thought, maybe that's wrong...

In which case did `,` as a separator come up as a problem? I guess with expressions, you could just parenthesize the arguments?

danebou:
The issue came up with adding another parameter to the for loop: `for (;; i++, j++)` : `for (;; i++PERM(,j++, )`

simonlindholm:
Ah, I see. Work-aroundable by `PERM_GENERAL(for (;; i++), for (;; i++, j++)`, but that's not very nice of course... One other option could be to accept a separator as an additional argument, e.g. `PERM_GENERAL[|](|i++)`... but that probably just complicates parsing...

Argh. I'd like say that we should just merge this PR and not overthink this, but `|` is super ugly, and does come up unparenthesized, e.g. with PERM_RANDOMIZE()s that cover large parts of the function.

danebou:
Yeah it’s ugly. On top of that for loop problem I had another perm in the initializer for the for loop, so the workaround becomes a lot worse. But given the or-ing I’m not quite satisfied. It’s way less conventional but what if we used multidim array like syntax: ‘PERM[arg1][arg2]’ or PERM([arg1][arg2])? I can’t think of a case where we would need to split an array lookup?

simonlindholm:
Another option: remove the outermost layer of parentheses from each argument (if there is one).

simonlindholm:
A nice thing about the current method is also that it matches how C macros behave, even including the rule about commas within parentheses not counting as argument separators. So we could e.g. introduce some perms that are easier to implement at the AST level as macros.

danebou:
That's a good point. Perhaps we just stick to commas, and make `(,)` and escape sequence for `,`? That way the AST wouldn't complain (I think)?

simonlindholm:
Oh yeah, `(,)` would work, that isn't used in C and thus less ambiguous than my idea. (`for (;; i++PERM((, j++), )` vs `for (;; i++PERM((,) j++, )`.)

PR #33, opened 2019-04-19
Comma escape sequence

danebou:


perm/perm_gen.py
```diff
@@ -49,6 +49,9 @@ def get_parenthesis_args(s: str) -> Tuple[List[str], str]:
                     break
             current += c
     assert level == 0, "Error, no closing parenthesis found"
+
+    # Comma escape sequence
+    args = [arg.replace('(,)', ',') for arg in args]
```
simonlindholm: I think we should do this when constructing TextPerm's. Otherwise you need to do `((,))` when you have nested perms.

PR #44, opened 2020-01-07
Add inequality perm to randomizer

Synray:


src/randomizer.py
```diff
@@ -815,6 +815,32 @@ def perm_reorder_stmts(
...
+    if not cands:
+        return False
+
+    node = random.choice(cands)
+
+    # Does not simplify, 'a <= (b + 1)' becomes 'a < ((b + 1) - 1)'
+
+    if len(node.op) == 1:
+        node.op += '='
+        node.right = ca.BinaryOp('+', node.right, ca.Constant('int', 1))
```
simonlindholm: this turns `a > b` into `a >= b + 1` which is good, but `a < b` into `a <= b + 1` which is less so. :)

Would be nice to randomize whether the LHS or RHS gets adjusted by 1, as well.

Synray:
Is is cleaner to return after `if random` or to use `else`?

simonlindholm:
About the same I'd said. A third version would be to use elif. Anyway, it doesn't matter, I'll merge.

PR #47, opened 2020-01-08
Add struct_ref perm to randomizer

### LINKED ISSUE #36: Swap `a + b` <-> `&a[b]`
in one case `a[b].c = d;` needed to be `(a + b)->c = d` across a range of statements

Synray:
This is pretty rough, but it successfully found (*((a + b) - 1)) which was needed for a match. 

Fixes #36.


src/randomizer.py
```diff
@@ -865,6 +865,110 @@ def minus1(node: ca.Node) -> ca.BinaryOp:
...
+    class Visitor(ca.NodeVisitor):
+        def visit_StructRef(self, node: ca.StructRef) -> None:
+            if region.contains_node(node):
+                cands.append(node)
+    Visitor().visit(fn.body)
+    if not cands:
+        return False
+
+    # TODO: This might need to handle nested BinaryOps
+    def choose_side(left: ca.Node, right: ca.BinaryOp) -> ca.BinaryOp:
```
simonlindholm: "randomize_associative_binop"? we should really have this as a standalone randomization pass, that transforms
a + (b ± c) <-> (a + b) ± c
a - (b ± c) <-> (a - b) ∓ c
but of course it doesn't hurt to do it here specifically as well.

This code as well some other seems to assume that + and - are the only BinaryOps that exist. Now, this is *that* big of a problem; if it does something weird with say multiplication we'll just get code that doesn't match very well and move on. But it might be worth fixing for sanity and to improve performance.

Synray: This is just a hack to properly convert array references, I _think_ it handles other binops on either side, since it uses the right side's operator and doesn't touch the left. I'll have to test it more.

Synray:
This is working now, and I'm fairly happy with the code even though I'm probably missing a clever pythonic way of simplifying it. I added the test file I used in case you want to test it yourself.
After you review it I'll remove the commented out prints.

src/randomizer.py
```diff
@@ -865,6 +865,186 @@ def minus1(node: ca.Node) -> ca.BinaryOp:
...
+        # This should be possible but it would mess up my crappy list handling code :|
+        #
+        #if isinstance(node, ca.UnaryOp) and node.op == '*':
+        #    return node.expr
+        return ca.UnaryOp('&', node)
+
+    def rec(node: ca.Node, nodes: List[ca.Node]) -> Tuple[bool, bool]:
+        """Add each child node of the StructRef to a list, stopping when the node is not a
+        dereference or address-of unary operator. Returns 'failed' when a UnaryOp with an
+        op other than * or & was encountered."""
```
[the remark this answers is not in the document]
Synray: Is it possible to use a NamedTuple here? Something like `-> NamedTuple[['binop_or_aref', bool], ['failed', bool]]`

src/randomizer.py
```diff
@@ -865,6 +865,186 @@ def minus1(node: ca.Node) -> ca.BinaryOp:
...
+        return False, True
+
+    # (Oh god why)
+    def apply_child(parent: ca.Node, func) -> None:
+        if isinstance(parent, ca.StructRef):
+            parent.name = func(parent.name)
+        elif isinstance(parent, ca.UnaryOp):
+            parent.expr = func(parent.expr)
+
+    def set_child(parent: ca.Node, child: ca.Node) -> None:
```
simonlindholm: same as `apply_child(parent, lambda _: child)` (moving this comment up here; I accidentally commented on the test before)

src/randomizer.py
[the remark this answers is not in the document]
simonlindholm: would also work, but exceptions are far cleaner IMO

src/randomizer.py
```diff
@@ -865,6 +865,186 @@ def minus1(node: ca.Node) -> ca.BinaryOp:
...
+    if struct_ref.type == '->':
+        struct_ref.type = '.'
+        struct_ref.name = deref(struct_ref.name)
+        nodes.append(struct_ref.name)
+        #print('-> changed: \033[32m',cg.visit(struct_ref),'\033[m')
+    nodes.append(struct_ref)
+    if not binop_or_array_ref:
+        if random.choice([True, False]):
+            apply_child(struct_ref, addr)
+            struct_ref.type = '->'
```
simonlindholm: since you know that this is a struct reference, you might as well avoid the unnecessarily generic apply_child

src/randomizer.py
```diff
@@ -865,6 +865,186 @@ def minus1(node: ca.Node) -> ca.BinaryOp:
...
+    #print(cg.visit(struct_ref),end='')
+    #print('\033[m')
+
+    # Step 1: Simplify (...)->c to (*(...)).c
+    if struct_ref.type == '->':
+        struct_ref.type = '.'
+        struct_ref.name = deref(struct_ref.name)
+        nodes.append(struct_ref.name)
+        #print('-> changed: \033[32m',cg.visit(struct_ref),'\033[m')
+    nodes.append(struct_ref)
```
simonlindholm: At this point and forward I completely lose track of what's going on. Why do we recurse arbitrarily deep but only use indices [0], [1] and [2]? Can we rewrite the recursion as a loop? Can we give [0], [1], [2] names? types? explain their nesting structure? either this needs more comments, or preferably rewriting to avoid the list thing.

test/test_struct_ref.py
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/env python
+
+from pycparser import c_ast as ca, c_parser, c_generator
+from pathlib import Path
+from random import Random
+from itertools import product
+import copy
+
+#TODO TODO TODO:
+#   * Make this a real unit test instead of copying perm_struct_ref :\
```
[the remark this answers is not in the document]
Synray: Removing the test for now until a more finalized version of this code is ready.

PR #55, opened 2020-02-18
Add perm_cast_simple

### LINKED ISSUE #31: Randomly insert casts

Synray:
Resolves #31

### DIFF
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -12,7 +12,7 @@ from pycparser import c_ast as ca, c_parser, c_generator
 from . import ast_util
 from .ast_util import Block, Indices, Statement, Expression
 from .ast_types import (SimpleType, TypeMap, build_typemap, decayed_expr_type,
-        resolve_typedefs, same_type, set_decl_name)
+        resolve_typedefs, same_type, allowed_simple_type, set_decl_name)
 
 # Set to true to perform expression type detection eagerly. This can help when
 # debugging crashes in the ast_types code.
@@ -119,7 +119,6 @@ def get_block_expressions(block: Block, region: Region) -> List[Expression]:
             replace_subexprs(stmt, visitor)
     rec(block)
     return exprs
-
 
 def compute_write_locations(
     top_node: ca.Node, indices: Indices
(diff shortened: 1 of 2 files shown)


src/randomizer.py
```diff
@@ -905,6 +905,52 @@ def visitor(expr: Expression) -> None:
...
+    expr = random.choice(cands)
+    type: SimpleType = decayed_expr_type(expr, typemap)
+    base_type = resolve_typedefs(type, typemap)
+
+    if not isinstance(base_type, ca.TypeDecl):
+        return False
+    if not isinstance(base_type.type, ca.IdentifierType):
+        return False
+    if all(x not in base_type.type.names for x in ['int', 'char', 'long', 'short', 'signed', 'unsigned', 'float', 'double']):
+        return False
```
simonlindholm: this pattern occurs in some other place as well, can we make a helper for it?

src/randomizer.py
```diff
@@ -905,6 +905,52 @@ def visitor(expr: Expression) -> None:
...
+
+    integral_type: List[List[str]] = [['int'], ['char'], ['long'], ['short'], ['long', 'long']]
+    floating_type: List[List[str]] = [['float'], ['double'], ['long', 'double']]
```
simonlindholm: let's remove long double, IDO doesn't support it, and it seems like it should be really easy to spot by hand.

src/randomizer.py
```diff
@@ -905,6 +905,52 @@ def visitor(expr: Expression) -> None:
...
+    new_type: List[str]
+    if random.choice([True, False]):
+        # Cast to integral type, sometimes unsigned
+        sign: List[str] = random.choice([[], ['unsigned']])
+        new_type = sign + random.choice(integral_type)
+    else:
+        # Cast to floating point type
+        new_type = random.choice(floating_type)
+
+    visit_replace(fn.body, lambda n, _: ca.Cast(ca.Typename(name=None,quals=[], type=ca.TypeDecl(None, [], ca.IdentifierType(new_type))), expr) if n is expr else None)
```
simonlindholm: pretty ugly one-liner; break out into a separate function and/or format with black.

src/randomizer.py
```diff
@@ -905,6 +905,52 @@ def visitor(expr: Expression) -> None:
...
+    # Find a random expression
+    def rec(block: Block) -> None:
+        for stmt in ast_util.get_block_stmts(block, False):
+            ast_util.for_nested_blocks(stmt, rec)
+            def visitor(expr: Expression) -> None:
+                if not region.contains_node(expr):
+                    return
+                cands.append(expr)
+            replace_subexprs(stmt, visitor)
+    rec(fn.body)
```
simonlindholm: would be nice to make this a helper function

Synray:
Also halved mask and cast_simple's probabilities to 10, they sometimes generate identical asm.

PR #58, opened 2020-03-07
Add a branch for ARM port

laqieer:


simonlindholm:
Nice! I can see three ways of managing this:
- add it to master, with compiler/architecture if statements
- keep it in a branch in this upstream repo, merging from master occasionally
- keep it in your fork, merging from master occasionally

I think for now the third option seems easiest, and would make it easy for you to add compiler-specific randomization passes. The first would also be reasonable, but needs more work from this PR if that's the way you want to go.

laqieer:
The only disadvantage of the third option is that a fork cannot show in GitHub's search results, which is acceptable considering it only aims to a few users. The first is best in the long run, making it easier to expand CPU architecture support.

simonlindholm:
I added a mention of your fork to the README, to improve searchability slightly: https://github.com/simonlindholm/decomp-permuter/commit/5cf1ddb87d71f35ebc7d793644c9864c877f52e9

PR #62, opened 2020-03-22
%lo / %hi should match any literal

### LINKED ISSUE #41: %lo / %hi should match any literal
The ground truth shouldn't need to use %lo/%hi. Currently %lo/%hi need to match completely or they give a super large penalty, which I suspect might be a huge footgun.

We should either rip out macro_arg from scorer.py, or define a custom equality operator which says that an empty macro_arg matches a non-empty one (seems better?).

ethteck:
Fixes #41

### DIFF
--- a/src/scorer.py
+++ b/src/scorer.py
@@ -13,7 +13,7 @@ from .objdump import objdump
 class DiffAsmLine:
     line: str = attr.ib(cmp=False)
     mnemonic: str = attr.ib()
-    macro_arg: str = attr.ib()
+    macro_arg: str = attr.ib(cmp=False)
 
     def __init__(self, line: str) -> None:
         self.line = line
@@ -58,10 +58,33 @@ class Scorer:
         deletions = []
         insertions = []
 
+        def lo_hi_match(old: str, new:str):
+            old_lo = old.find("%lo")
+            old_hi = old.find("%hi")
+            new_lo = new.find("%lo")
+            new_hi = new.find("%hi")
+
+            if (old_lo != -1 and new_lo != -1):
+                old_idx = old_lo
+                new_idx = new_lo
+            elif (old_hi != -1 and new_hi != -1):
+                old_idx = old_hi
+                new_idx = new_hi
+            else:
+                return False
+
+            old_inner = old[old_idx + 4 : -1]
+            new_inner = new[new_idx + 4 : -1]
+            return old_inner == ".rodata" or new_inner == ".rodata"
+
         def diff_sameline(old: str, new: str) -> None:
             nonlocal score
             if old == new:
                 return
+
+            if lo_hi_match(old, new):
+                return
+
             # Probably regalloc difference, or signed vs unsigned
             score += self.PENALTY_REGALLOC
 
--- a/src/scorer.py
+++ b/src/scorer.py
@@ -53,16 +53,16 @@ class Scorer:
         deletions = []
         insertions = []
 
-        def lo_hi_match(old: str, new:str):
+        def lo_hi_match(old: str, new: str):
             old_lo = old.find("%lo")
             old_hi = old.find("%hi")
             new_lo = new.find("%lo")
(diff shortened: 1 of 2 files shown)


src/scorer.py
```diff
@@ -13,7 +13,7 @@
 class DiffAsmLine:
     line: str = attr.ib(cmp=False)
     mnemonic: str = attr.ib()
-    macro_arg: str = attr.ib()
+    macro_arg: str = attr.ib(cmp=False)
```
simonlindholm: should be possible to remove `macro_arg` completely now, actually, it was only there for SequenceMatcher's sake

src/scorer.py
```diff
@@ -58,10 +58,33 @@ def score(self, cand_o: Optional[str]) -> Tuple[int, str]:
...
+                new_idx = new_lo
+            elif (old_hi != -1 and new_hi != -1):
+                old_idx = old_hi
+                new_idx = new_hi
+            else:
+                return False
+
+            old_inner = old[old_idx + 4 : -1]
+            new_inner = new[new_idx + 4 : -1]
+            return old_inner == ".rodata" or new_inner == ".rodata"
```
simonlindholm: We should also allow `%lo(.rodata + 0x10)` and `%lo(.data + ...)` -- how about `old_inner.startswith(".") or new_inner.startswith(".")`?

simonlindholm: Compare the parts before old_idx and new_idx as well, they may contain regalloc diffs

ethteck: as for the parts before old_idx and new_idx, what should the behavior be if they're the same vs different?

ethteck:
Hm... mypy is reporting tons of errors and issues, many of which seem to be unrelated to my changes. I am unable to find "black". And the test suite fails instantly...

```
E/usr/lib/python3.6/tempfile.py:939: ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/t
mph67libr6'>
  _warnings.warn(warn_message, ResourceWarning)

======================================================================
ERROR: setUpClass (test_perm.TestStringMethods)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ethteck/repos/decomp-permuter/test/test_perm.py", line 44, in setUpClass
    shutil.copy2(target_o, file_target)
  File "/usr/lib/python3.6/shutil.py", line 263, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.6/shutil.py", line 103, in copyfile
    if _samefile(src, dst):
  File "/usr/lib/python3.6/shutil.py", line 88, in _samefile
    return os.path.samefile(src, dst)
  File "/usr/lib/python3.6/genericpath.py", line 96, in samefile
    s1 = os.stat(f1)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType

----------------------------------------------------------------------
Ran 0 tests in 0.036s

FAILED (errors=1)
```

simonlindholm:
Hm, weird. mypy for me claims
```
$ mypy permuter.py 
src/scorer.py:55: error: Function is missing a return type annotation
```
(with mypy --version saying 0.660), and the tests pass with your changes... I'm using Python 3.6.9 on Ubuntu in case it matters. Is there a message before the error? Seems like target_o is None, and thus the compile failed. `black` you will indeed have to install manually.

Don't worry too much about this though, I'm fine with merging the PR regardless.

src/scorer.py
simonlindholm: I think it would be fine to just return false here if they are different, thus resulting in PENALTY_REGALLOC if they differ (and else score 0).

ethteck:
My mypi version is 0.560-1 - I'd have to look elsewhere for a newer version. Maybe that's the problem. I got black to work, but it suggested a ton of changes in files I didn't touch so I only committed the scorer.py changes. Let me know if these changes work!

simonlindholm:
Looks good!

PR #68, opened 2020-03-25
Adding flag for reporting better scores only

ethteck:


src/main.py
```diff
@@ -278,7 +284,7 @@ def post_score(context: EvalContext, permuter: Permuter, result: EvalResult) ->
             print(
                 f"[{permuter.unique_name}] found a better score! ({score_value} vs {permuter.base_score})"
             )
-        else:
+        elif not context.options.better_only:
```
simonlindholm: this check is unnecessary given the one above

ethteck:
@simonlindholm sorry this took me 4 months, but I just updated it! Let me know what you think.

simonlindholm:
Looks good!

PR #72, opened 2020-04-10
Add option to use sp differences when scoring

Synray:


src/objdump.py
```diff
@@ -72,7 +72,7 @@ def parse_relocated_line(line: str) -> Tuple[str, str, str]:
     return before, imm, after
 
 
-def simplify_objdump(input_lines: List[str]) -> List[str]:
+def simplify_objdump(input_lines: List[str], stack_differences: bool) -> List[str]:
```
simonlindholm, line 75: In general, bool parameters are very hard to understand what they do when you read the code at a call site. Prefer either passing in the entire Options object or a subobject thereof that deals only with scoring, or making this a keyword-only param:
```
def simplify_objdump(input_lines: List[str], *, stack_differences: bool) -> List[str]:
    ...
...
simplify_objdump(lines, stack_differences=True or False)
```

src/scorer.py
```diff
@@ -6,7 +6,7 @@
 
 import attr
 
-from .objdump import objdump
+from .objdump import objdump, sprel
```
simonlindholm, line 9: name sprel something clearer now that it's imported from elsewhere

src/scorer.py
```diff
@@ -83,6 +84,15 @@ def diff_sameline(old: str, new: str) -> None:
             if lo_hi_match(old, new):
                 return
 
+            if self.stack_differences:
+                oldsp = re.search(sprel, old)
+                newsp = re.search(sprel, new)
+                if oldsp and newsp:
+                    oldrel = int(oldsp.group(1), 0)
+                    newrel = int(newsp.group(1), 0)
+                    score += abs(oldrel - newrel)
```
simonlindholm, line 93: multiply by `PENALTY_STACKDIFF` for uniformity's sake (which can be set to 1)

src/main.py
```diff
@@ -410,7 +411,7 @@ def run_inner(options: Options, heartbeat: Callable[[], None]) -> List[int]:
             print(base_c)
 
         compiler = Compiler(compile_cmd, options.show_errors)
-        scorer = Scorer(target_o)
+        scorer = Scorer(target_o, context.options.stack_differences)
```
simonlindholm, line 414: context.options -> options

simonlindholm:
cc @pixel-stuck who might have opinions on this

PR #81, opened 2020-06-26
Add perm_split_assignment

Synray:
Not sure how this will handle for loops, assignments inside expressions, etc.

### DIFF
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -1412,6 +1412,57 @@ def perm_struct_ref(
     ensure(changed)
 
 
+def perm_split_assignment(
+    fn: ca.FuncDef, ast: ca.FileAST, indices: Indices, region: Region, random: Random
+) -> bool:
+    """Split assignments of the form a (.)= b . c . d ...; into a = b; a (.)= c . d ...;, a = b . c; a (.)= d ...; etc."""
+    cands = []
+    # Look for assignments of the form 'var (.)= binaryOp'
+    class Visitor(ca.NodeVisitor):
+        def visit_Assignment(self, node: ca.Assignment) -> None:
+            if isinstance(node.rvalue, ca.BinaryOp) and region.contains_node(node):
+                cands.append(node)
+
+    Visitor().visit(fn.body)
+    ensure(cands)
+
+    assign = random.choice(cands)
+    var = assign.lvalue
+
+    binops = []
+
+    def collect_binops(node: ca.BinaryOp) -> None:
+        nonlocal binops
+        if isinstance(node.left, ca.BinaryOp):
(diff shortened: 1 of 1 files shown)


src/randomizer.py
```diff
@@ -1412,6 +1412,57 @@ def get_child(parent: Union[ca.StructRef, ca.UnaryOp]) -> ca.Node:
...
+    Visitor().visit(fn.body)
+    ensure(cands)
+
+    assign = random.choice(cands)
+    var = assign.lvalue
+
+    binops = []
+
+    def collect_binops(node: ca.BinaryOp) -> None:
+        nonlocal binops
```
simonlindholm, line 1435: fwiw this can be removed; nonlocal is only needed when you need to assign to an outer variable, not when poking at it using member functions

src/randomizer.py
```diff
@@ -1412,6 +1412,57 @@ def get_child(parent: Union[ca.StructRef, ca.UnaryOp]) -> ca.Node:
...
+        side = split.left
+        split.left = copy.deepcopy(var)
+    else:
+        side = split.right
+        split.right = copy.deepcopy(var)
+
+    new_assign = ca.Assignment("=", copy.deepcopy(var), side)
+
+    ins_cands = get_insertion_points(fn, region)
+    ensure(ins_cands)
```
simonlindholm, line 1457: Avoid calling `ensure` after modifying the AST, since it will cause some other pass to operate on a damaged AST in a potentially hard to debug manner. `assert` would work I think (i.e. assert False just after the loop below), but it may be safer to move the ins_cands creation and loop through ins_cands to just after `assign` is assigned.

src/randomizer.py
```diff
@@ -1429,10 +1433,18 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
...
     var = assign.lvalue
 
+    ins_cands = get_insertion_points(fn, region)
+    ensure(ins_cands)
+
+    for ins_block, ins_index, node in ins_cands:
+        if node is assign:
+            break
+    else:
+        assert False, "original assign should always be found in insertion points"
```
simonlindholm, line 1443: either replace by `raise RandomizationFailure`, or move this down back again (since all ensures are gone, which are the only reason to keep this up here)

src/randomizer.py
```diff
@@ -1429,10 +1433,18 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
     assign = random.choice(cands)
     var = assign.lvalue
 
+    ins_cands = get_insertion_points(fn, region)
+    ensure(ins_cands)
```
simonlindholm, line 1437: remove the ensure

PR #82, opened 2020-07-31
Modified compile command generation for PM support

ethteck:
There are probably more pythonic ways to do what I've done here, so I'm open to any suggestions if you'd like something to be written differently.

### DIFF
--- a/import.py
+++ b/import.py
@@ -39,7 +39,7 @@ SETTINGS_FILES = ["permuter_settings.toml", "tools/permuter_settings.toml"]
 
 
 def formatcmd(cmdline: List[str]) -> str:
-    return " ".join(shlex.quote(arg) for arg in cmdline)
+    return " ".join(arg if arg == '|' else shlex.quote(arg) for arg in cmdline)
 
 
 def parse_asm(asm_file: str) -> Tuple[str, str]:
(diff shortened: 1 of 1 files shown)


import.py
```diff
@@ -400,13 +400,19 @@ def import_c_file(
...
+
+
 def write_compile_command(compiler: List[str], cwd: str, out_file: str) -> None:
     with open(out_file, "w", encoding="utf-8") as f:
         f.write("#!/usr/bin/env bash\n")
         f.write('INPUT="$(readlink -f "$1")"\n')
         f.write('OUTPUT="$(readlink -f "$3")"\n')
         f.write(f"cd {shlex.quote(cwd)}\n")
-        f.write(formatcmd(compiler) + ' "$INPUT" -o "$OUTPUT"\n')
+        f.write(finalize_compile_command(formatcmd(compiler)))
```
simonlindholm, line 415: I would prefer not to modify `formatcmd` since that modification is wrong in other situations (which use subprocess.run rather than outputting a shell script), and since post-processing its result may in theory do stuff to quoted `|`'s. How about this instead:
```py
def finalize_compile_command(cmdline):
    quoted = [arg if arg == "|" else shlex.quote(arg) for arg in cmdline]
    try:
        ind = quoted.index("|")
    except ValueError:
        ind = len(quoted)
    return " ".join(quoted[:ind] + ['"$INPUT"'] + quoted[ind:] + ["-o", '"$OUTPUT"'])
```

ethteck: Sounds good! Pushed the changes.

PR #85, opened 2020-09-14
add float literal randomization

krimtonz:
randomizes floating point literals
0.x -> .x
x.0 -> x.
using or not using f to indicate float


simonlindholm (approved):
Looks great!

PR #87, opened 2020-09-21
Added perm_equals_operator pass

ethteck:
Some notes:
* I'm not in love with the name I picked. Let me know if you think it should change.
* I gave the weight for it `5` - not sure if that should change

### DIFF
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -1137,6 +1137,7 @@ def perm_equals_operator(
         def visit_Assignment(self, node: ca.Assignment) -> None:
             if node.op == "=" and \
             region.contains_node(node) and \
+            isinstance(node.rvalue, ca.BinaryOp) and \
             ast_util.equal_ast(node.lvalue, node.rvalue.left) and \
             node.rvalue.op in operators:
                 cands.append(node)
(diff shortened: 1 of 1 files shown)


src/randomizer.py
```diff
@@ -1126,6 +1126,31 @@ def perm_reorder_stmts(
...
+    cands: List[ca.Assignment] = []
+    operators = ["+", "-", "*", "/", "<<", ">>", "^", "|", "&"]
+
+    class Visitor(ca.NodeVisitor):
+        def visit_Assignment(self, node: ca.Assignment) -> None:
+            if node.op == "=" and \
+            region.contains_node(node) and \
+            isinstance(node.rvalue, ca.BinaryOp) and \
+            ast_util.equal_ast(node.lvalue, node.rvalue.left) and \
+            node.rvalue.op in operators:
```
simonlindholm, line 1142:
```suggestion
            if (node.op == "=" and
            region.contains_node(node) and
            isinstance(node.rvalue, ca.BinaryOp) and
            ast_util.equal_ast(node.lvalue, node.rvalue.left) and
            node.rvalue.op in operators):
```
and then let `black` auto-format this

simonlindholm (approved):
Looks good! perm_compound_assignment would be a better name. It would also be nice to have a pass for the inverse transformation, but that can be separate.

PR #89, opened 2020-11-19
Add --best-only

onlymx13:
By analogy with --better-only


onlymx13:
This was black-formatted.

PR #90, opened 2020-12-15
Add remove_ast and duplicate_assignment

Synray:


src/randomizer.py
```diff
@@ -1745,6 +1745,77 @@ def collect_binops(node: ca.BinaryOp) -> None:
...
+        expr = cand.expr
+    elif isinstance(cand, ca.BinaryOp):
+        if isinstance(cand.left, ca.Constant) and isinstance(cand.right, ca.Constant):
+            expr = random.choice([cand.left, cand.right])
+        elif isinstance(cand.left, ca.Constant):
+            expr = cand.right
+        else:
+            expr = cand.left
+    elif isinstance(cand, ca.If):
+        expr = cand.iftrue
```
simonlindholm, line 1789: Instead of isinstance checks here, we could have `cands` contain tuples of "(expr, expr to replace it with)" and move code into the visitor.

This would also simplify BinaryOp since it can add two entries into cands if both lhs and rhs are literals.

PR #92, opened 2020-12-22
Add perm_decl

Synray:


src/randomizer.py
```diff
@@ -415,7 +415,9 @@ def random_type(random: Random) -> SimpleType:
     new_names: List[str] = []
     if random_bool(random, 0.5):
         new_names.append("unsigned")
-    new_names.append(random.choice(["char", "short", "int", "int"]))
+    new_names.append(
+        random.choice(["char", "short", "int", "int", "long", "long long"])
+    )
```
simonlindholm, line 420:
```suggestion
    new_names.extend(
        random.choice(["char", "short", "int", "int", "long", "long long"]).split()
    )
```

src/randomizer.py
```diff
@@ -1815,6 +1817,29 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
     ast_util.insert_statement(tob, toi, dup)
 
 
+def perm_decl(
+    fn: ca.FuncDef, ast: ca.FileAST, indices: Indices, region: Region, random: Random
+) -> None:
```
simonlindholm, line 1822: add a doc comment

src/randomizer.py
```diff
@@ -1815,6 +1817,29 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
...
+    vars: List[str] = []
+
+    class Visitor(ca.NodeVisitor):
+        def visit_Decl(self, decl: ca.Decl) -> None:
+            if decl.name:
+                vars.append(decl.name)
+
+    Visitor().visit(fn.body)
+    ensure(vars)
```
simonlindholm, line 1831: unnecessary, I suppose

src/randomizer.py
```diff
@@ -1815,6 +1817,29 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
     ast_util.insert_statement(tob, toi, dup)
 
 
+def perm_decl(
```
simonlindholm, line 1820: perm_new_unused_decl would be more descriptive

src/randomizer.py
```diff
@@ -415,7 +415,9 @@ def random_type(random: Random) -> SimpleType:
     new_names: List[str] = []
     if random_bool(random, 0.5):
         new_names.append("unsigned")
-    new_names.append(random.choice(["char", "short", "int", "int"]))
+    new_names.extend(
+        random.choice(["char", "short", "int", "int", "long", "long long"]).split()
```
simonlindholm, line 419: hm, maybe this gives a bit too much weight to long and long long?
how about
```suggestion
        random_weighted(random, [
            (["char"], 1),
            (["short"], 1),
            (["int"], 2),
            (["long"], 0.5),
            (["long", "long"], 0.5),
        ])
```

PR #93, opened 2021-01-09
Sometimes add multiple masks in perm_add_mask

Synray:
Since I've seen regalloc cycle with 3 or 4 masks, I just chose 12 arbitrarily since 12 has a lot of factors.

### DIFF
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -1448,7 +1448,7 @@ def perm_add_mask(
     mask = random.choice(masks) + random.choice(["", "u"])
 
     new_expr = ca.BinaryOp("&", expr, ca.Constant("int", mask))
-    if random_bool(random, 0.5):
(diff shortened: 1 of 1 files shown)


src/randomizer.py
```diff
@@ -1447,6 +1448,10 @@ def perm_add_mask(
     mask = random.choice(masks) + random.choice(["", "u"])
 
     new_expr = ca.BinaryOp("&", expr, ca.Constant("int", mask))
+    if random_bool(random, 0.5):
```
simonlindholm, line 1451: 0.5 feels a tad high to me for something which is certainly a fakematch, but at the same time it's just a 1/12 chance the number of bitands will be correct... hm. Maybe lower to 0.3 and also raise perm_add_mask weight 10->15?

simonlindholm (approved):

PR #94, opened 2021-01-15
Ninja support and new build_system setting

ethteck:


import.py
```diff
@@ -540,16 +552,27 @@ def main() -> None:
     )
     args = parser.parse_args()
 
-    make_flags = args.make_flags + ["PERMUTER=1"]
+    settings: Mapping[str, object] = {}
+    settings_dir = find_root_dir(args.c_file, SETTINGS_FILES)
```
simonlindholm, line 556: Maybe add `["Makefile", "makefile", "build.ninja"]` to the list of searched files, and remove the find_root_dir calls in find_build_command_line? Feels a bit convoluted to search for root dir twice.

import.py
```diff
@@ -111,19 +111,16 @@ def create_directory(func_name: str) -> str:
             pass
 
 
-def find_makefile_dir(filename: str) -> str:
+def find_root_dir(filename: str, pattern: List[str]) -> str:
```
simonlindholm, line 114: Optional[str] (can you run mypy?)

import.py
```diff
@@ -111,19 +111,16 @@ def create_directory(func_name: str) -> str:
...
     while dirname and (not old_dirname or len(dirname) < len(old_dirname)):
-        for fname in ["makefile", "Makefile"]:
+        for fname in pattern:
             if os.path.isfile(os.path.join(dirname, fname)):
                 return dirname
         old_dirname = dirname
         dirname = os.path.dirname(dirname)
 
-    print(f"Missing makefile for file {filename}!", file=sys.stderr)
-    sys.exit(1)
```
simonlindholm, line 125: `return None` (or reinstate the error in this function, either way)

import.py
```diff
@@ -162,16 +159,23 @@ def fixup_build_command(
 
 
 def find_build_command_line(
-    c_file: str, make_flags: List[str]
+    root_dir: str, c_file: str, make_flags: List[str], build_system: str
 ) -> Tuple[List[str], List[str], str]:
```
simonlindholm, line 163:
```suggestion
) -> Tuple[List[str], List[str]]:
```

import.py
```diff
@@ -204,7 +217,7 @@ def find_build_command_line(
             else ""
         )
         print(
-            "Failed to find compile command from makefile output. "
+            "Failed to find compile command from build script output. "
             f"Please ensure 'make -Bn --debug=j {formatcmd(make_flags)}' "
```
simonlindholm, line 221: this will be misleading with ninja, maybe worth fixing

import.py
```diff
@@ -215,28 +228,21 @@ def find_build_command_line(
...
-    return output[0], assembler, makefile_dir
+    return output[0], assembler
 
 
 PreserveMacros = Tuple[Pattern[str], Callable[[str], str]]
 
 
 def build_preserve_macros(
-    cwd: str, preserve_regex: Optional[str]
+    cwd: str, preserve_regex: Optional[str], data: Mapping[str, object]
```
simonlindholm, line 244: data -> settings

nanaian:
Resolved the review comments, and `mypy` now outputs only the two errors that already exist on `master`.

import.py
```diff
@@ -227,9 +230,9 @@ def find_build_command_line(
     if len(output) > 1:
         output_lines = "\n".join(map(formatcmd, output))
         print(
-            f"Error: found multiple compile commands for {rel_c_file}:\n{output_lines}\n"
-            "Please modify the build script such that if PERMUTER = 1, "
-            "only a single compile command is included.",
+            f"Error: found multiple compile commands for {rel_c_file}:\n{output_lines}\n{close_extra}"
```
simonlindholm, line 233:
```suggestion
            f"Error: found multiple compile commands for {rel_c_file}:\n{output_lines}\n"
```
(close_extra isn't defined here, nor would its definition from the if block above make sense here)

simonlindholm (approved):

PR #105, opened 2021-11-08
Updates to accommodate pycparser update

ethteck:


simonlindholm:
Can you update stubs/ too similar to how m2c did it, and make sure mypy runs without errors?

stubs/pycparser/c_ast.py
```diff
@@ -344,20 +360,22 @@ def __init__(self, coord: Optional[Coord] = None):
...
 
     def __init__(
         self,
         name: Optional[str],
         quals: List[str],
+        align: List[Alignas],
         storage: List[str],
         funcspec: List[str],
-        type: Union_[Type, "Struct", "Union", "Enum"],
-        init: Optional[Union_[Expression, "InitList"]],
```
simonlindholm, line 360: revert these changes (these fixes to Decl should really be mirrored to m2c, I must have forgot)

ethteck: just these two lines, or all changes to the Decl class?

simonlindholm: align is still needed, but revert the type and init changes, here and the `type` thing higher up

PR #106, opened 2022-01-25
Add Dockerfile for pah

mkst:
This is my interpretation of `simonlindholm/ido:4.2` but it's ~~untested...~~ only tested against ido 5.3 currently

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,51 @@
+FROM python:3.9-slim as compilers
+
+RUN apt-get update && apt-get install -y curl wget
+
+RUN mkdir -p "/ido5.3" && \
+    curl -L "https://github.com/ethteck/ido-static-recomp/releases/download/master/ido-5.3-recomp-ubuntu-latest.tar.gz" | \
+    tar zx -C "/ido5.3"
+
+RUN mkdir -p "/ido7.1" && \
+    curl -L "https://github.com/ethteck/ido-static-recomp/releases/download/master/ido-7.1-recomp-ubuntu-latest.tar.gz" | \
+    tar zx -C "/ido7.1"
+
+RUN mkdir -p "/gcc2.8.1" && \
+    curl -L "https://github.com/pmret/gcc-papermario/releases/download/master/linux.tar.gz" | \
+    tar zx -C "/gcc2.8.1" && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN mkdir -p "/gcc2.8.1" && \
+    curl -L "https://github.com/pmret/gcc-papermario/releases/download/master/linux.tar.gz" | \
+    tar zx -C "/gcc2.8.1" && \
+    curl -L "https://github.com/pmret/binutils-papermario/releases/download/master/linux.tar.gz" | \
+    tar zx -C "gcc2.8.1"
+
+RUN mkdir -p /nu64 && \
+    wget -P /nu64 "https://github.com/Rainchus/JPDecompPracticeROM/blob/main/tools/build/linux/cc1" && \
+    wget -P /nu64 "https://github.com/Rainchus/JPDecompPracticeROM/blob/main/tools/build/linux/mips-nintendo-nu64-as"
```
simonlindholm, line 21: These github links feel like they could disappear at any moment... Would be nice if we had a custom repo for hosting the contents. Then the Dockerfile could be in the same repo. This also nicely avoids embedding a single dockerfile into the permuter repo: the vision is that separate pah networks could use different dockerfiles.

(If we do keep it in this repo the Dockerfile should move to some subdirectory, like src/net/controller/)

mkst: sure, shall I chat to @ethteck about a repo within decompals maybe? I think that we should pull binaries that we can.. but yeh these ones don't feel particularly stable -- I just wanted to add them as they are in the original image...

I *could* do 
```COPY --from=simonlindholm/ido:4.2```

for those binaries.. but it's bootstrapping... so not super clean.

simonlindholm: Sure, decompals sounds reasonable. I agree `COPY --from=simonlindholm/ido:4.2` isn't great, it's better if we have upstream sources listed so we can easily update, and I think Docker Hub expires images after a while too.

Not sure what the best tradeoff between not having builds break at random while also having upstream sources referenced is... Vendoring scripts that can be run manually and downloads stuff to some directory in the repo? Specifying commit hashes in the github links in the dockerfile and hoping repos vanishing is rare enough that breakage can be dealt with manually?

mkst:
Superseded by https://github.com/decompals/pah-docker/pull/1

PR #108, opened 2022-03-30
Add prelude.inc with support for abi fpr names

EllipticEllipsis:


prelude.inc
```diff
@@ -0,0 +1,43 @@
+.set noat
+.set noreorder
+.set gp=64
+.macro glabel label
+    .global \label
+    \label:
+.endm
+
+
+# Float register aliases (o32 ABI, odd ones are rarely used)
```
simonlindholm, line 10:
```suggestion
# Float register aliases (o32 ABI)
```

simonlindholm (approved):

PR #109, opened 2022-05-05
WIP MWCC support

snuffysasa:
Please read mwcc_usage.md.

This is very much a WIP / barebones setup I used for helping me match functions in melee with mwcc 1.2.5(e).

I was requested to PR this to this repo, however, in it's current state it definitely should not be merged into main.  It could go to a new branch or left as WIP.

I don't intend to work on this a ton more, however, if there is interest and I am asked / requested to, I would help with some requests to make this useful for others.


simonlindholm:
Thanks! I agree with not merging this as is; there's no reason why PPC should need a different setup from MIPS. I guess I'll leave this PR open for people to make use of until we get proper support (which I do believe should be fairly simple and require fewer changes than this PR makes).

snuffysasa:
closing this one

PR #111, opened 2022-05-26
default to selecting the last function in file when there is more than one

snuffysasa:
Please consider this PR.

A common workflow with mwcc ppc projects is having a very large context that is copied across scratches that often times has several inlines and nested inlines, these count as functions making therefore more than one function in the file, and the function name must be specified function.txt, but that is another step which makes slows down work flow.  Adding this PR would save time and make it easier to use by just defaulting to the last function in the file.  (since inlines are typically above in context)

So with this PR, you can still specify a function with that txt file, but in the case where its not specified and there are more than one function, it will just default to choosing the last (bottom) function.

### DIFF
--- a/src/permuter.py
+++ b/src/permuter.py
@@ -100,12 +100,8 @@ class Permuter:
             fns = _find_fns(source)
             if len(fns) == 0:
                 raise Exception(f"{self.source_file} does not contain any function!")
-            if len(fns) > 1:
-                raise Exception(
-                    f"{self.source_file} must contain only one function, "
-                    "or have a function.txt next to it with a function name."
-                )
-            self.fn_name = fns[0]
+            self.fn_name = fns[-1]
+            print("selecting function: ", self.fn_name)
         else:
             self.fn_name = fn_name
         self.unique_name = self.fn_name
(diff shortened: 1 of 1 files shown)


src/permuter.py
```diff
@@ -100,12 +100,8 @@ def __init__(
...
             if len(fns) == 0:
                 raise Exception(f"{self.source_file} does not contain any function!")
-            if len(fns) > 1:
-                raise Exception(
-                    f"{self.source_file} must contain only one function, "
-                    "or have a function.txt next to it with a function name."
-                )
-            self.fn_name = fns[0]
+            self.fn_name = fns[-1]
+            print("selecting function: ", self.fn_name)
```
simonlindholm, line 104:
```suggestion
            print("Selecting function: ", self.fn_name)
```
(would be nicer to have print() only happen in main.py, but we're already inconsistent about that)

simonlindholm (approved):
Fine with me, though I would encourage using an import script that creates function.txt. Note the comment about this code path being kinda legacy; it uses a regex to parse C and that regex does regularly fall down for complicated function definitions.

PR #112, opened 2022-05-26
trim source output to only show the target function code and below

snuffysasa:
Often times context is thousands of lines long (atleast for decomp projects I am involved in) and it is a pain to scroll down all the time to view the function of interest at the bottom, and also lots of unnecessary disk usage.

This simply finds the first instance of the function name and then only saves text after that point.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -63,6 +63,7 @@ class Options:
     use_network: bool = False
     network_debug: bool = False
     network_priority: float = 1.0
+    no_context_output: bool = False
 
 
 def restricted_float(lo: float, hi: float) -> Callable[[str], float]:
@@ -91,7 +92,7 @@ class EvalContext:
     permuters: List[Permuter] = field(default_factory=list)
 
 
-def write_candidate(perm: Permuter, result: CandidateResult) -> None:
+def write_candidate(perm: Permuter, result: CandidateResult, no_context_output: bool) -> None:
     """Write the candidate's C source and score to the next output directory"""
     ctr = 0
     while True:
@@ -104,11 +105,14 @@ def write_candidate(perm: Permuter, result: CandidateResult) -> None:
             pass
     source = result.source
 
-    fn_str_index = source.find(perm.fn_name)
 
     assert source is not None, "Permuter._need_to_send_source is wrong!"
     with open(os.path.join(output_dir, "source.c"), "x", encoding="utf-8") as f:
-        f.write(source[fn_str_index:])
+        if no_context_output:
+            fn_str_index = source.find(perm.fn_name)
+            f.write(source[fn_str_index:])
+        else:
+            f.write(source)
     with open(os.path.join(output_dir, "score.txt"), "x", encoding="utf-8") as f:
         f.write(f"{result.score}\n")
     with open(os.path.join(output_dir, "diff.txt"), "x", encoding="utf-8") as f:
@@ -175,7 +179,7 @@ def post_score(
             color = "\u001b[33m"
             msg = f"found different asm with same score ({score_value})"
         context.printer.print(msg, permuter, who, color=color)
-        write_candidate(permuter, result)
+        write_candidate(permuter, result, context.options.no_context_output)
     if not context.options.quiet:
         context.printer.progress(status_line)
     return score_value == 0
@@ -623,6 +627,13 @@ def main() -> None:
             Each server runs with a priority threshold, which defaults to 0.1,
             below which they will not run permuter jobs at all.""",
     )
+    parser.add_argument(
+        "--no-context-output",
+        dest="no_context_output",
+        default=False,
+        action="store_true",
+        help="Removes all text above the target function in the output source files",
+    )
 
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


simonlindholm:
Hm, I'm torn about this. It's not entirely unreasonable... I usually go entirely by diff.txt, but this could make source.c more practically useful.

There is a risk of false positives when naively searching for the first occurrence, for example it might find a declaration. Also, the permuter is able to make changes outside the function, e.g. change return types of referenced functions and signedness of globals -- these would then only show up in diff.txt.

If you use import.py it would strip your context down to only the functions/globals that are referenced, which makes large contexts much less of a problem, as well as improving compile speeds.

snuffysasa:
Ahh I didn't consider diff.txt.  But I still am interested in doing something here, maybe we can find a middle ground.

I'll explain. I understand that import.py will strip down the context .. however at-least for functions in ssbm(melee) you'll still end up needing to scroll down hundreds of lines even after being stripped down to just what is necessary.

I understand there is a risk of false positives, but in the case of a false positive it would only log more data(no harm imo), and especially considering the current implementation already just logs all.  And then in the cases of FP the user could decide whether they want to fix or leave it.

And then for diff.txt, I didn't consider this, however, when I am using this, I find want to just quickly copy paste the new candidate it found into decomp.me to evaluate. And it's harder to do that quickly with diff.txt.

Okay so some solutions I am thinking of, we could make the logger output an addition file, like diff.txt, base_function, plus ctx.h.  Or alternatively, I could set this up behind an optional python flag/param.

What do you think ?

simonlindholm:
Splitting source.c into source.c+ctx.c sounds like a good compromise, if we're able to do that. A bit tricky to do it accurately and with no perf penalty though -- CandidateResult.source would need to be split up and computed lazily and p@h would need to grow a "context" field. (Can't do it reliably on the client because the source may not be parseable with pycparser at that point due to macros.)

I'd be fine with an optional cmdline flag for applying your cut-off-context heuristic I think. Could expand it backward to the beginning of line to include the return type, btw.

src/main.py
```diff
@@ -103,9 +104,20 @@ def write_candidate(perm: Permuter, result: CandidateResult) -> None:
...
+
     assert source is not None, "Permuter._need_to_send_source is wrong!"
     with open(os.path.join(output_dir, "source.c"), "x", encoding="utf-8") as f:
-        f.write(source)
+        if no_context_output:
+            fn_str_index = source.find(perm.fn_name)
+            for i in range(fn_str_index-30, fn_str_index-1):
+                if (source[i] == '\n' or source[i] == '\r'):
+                    fn_str_index = i
+                    break
```
simonlindholm, line 116: extract this logic into a separate function, and use `rfind("\n")` to find the newline backward

snuffysasa:
> I'd be fine with an optional cmdline flag for applying your cut-off-context heuristic I think. Could expand it backward to the beginning of line to include the return type, btw.

I updated this PR to put the feature behind a cmd line flag.

src/main.py
```diff
@@ -103,9 +104,20 @@ def write_candidate(perm: Permuter, result: CandidateResult) -> None:
         except FileExistsError:
             pass
     source = result.source
+
+
     assert source is not None, "Permuter._need_to_send_source is wrong!"
     with open(os.path.join(output_dir, "source.c"), "x", encoding="utf-8") as f:
-        f.write(source)
+        if no_context_output:
+            fn_str_index = source.find(perm.fn_name)
```
simonlindholm, line 112: need to handle the case when find returns -1

src/main.py
```diff
@@ -621,6 +632,13 @@ def main() -> None:
             Each server runs with a priority threshold, which defaults to 0.1,
             below which they will not run permuter jobs at all.""",
     )
+    parser.add_argument(
+        "--no-context-output",
+        dest="no_context_output",
+        default=False,
```
simonlindholm, line 638: (this happens automatically when `action="store_true"`)

simonlindholm (commented):
might want to set up `black` for formatting if you're going to make many changes btw, see end of readme

snuffysasa:
I got black formatter setup and formatted this file with black, and completed the changes you asked for.

src/main.py
```diff
@@ -91,7 +92,18 @@ class EvalContext:
     permuters: List[Permuter] = field(default_factory=list)
 
 
-def write_candidate(perm: Permuter, result: CandidateResult) -> None:
+def trim_source(source, fn_name):
```
simonlindholm, line 95:
```suggestion
def trim_source(source: str, fn_name: str) -> str:
```
to avoid mypy complaints

src/main.py
```diff
@@ -621,6 +636,12 @@ def main() -> None:
             Each server runs with a priority threshold, which defaults to 0.1,
             below which they will not run permuter jobs at all.""",
     )
+    parser.add_argument(
+        "--no-context-output",
```
simonlindholm, line 640: (not in love with this flag name but I don't have a better suggestion off hand)

simonlindholm (approved):

PR #114, opened 2022-05-27
init attempt to fix r2 reloc for ppc, 

snuffysasa:
for loading external data with lfs and probably more.

### DIFF
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -159,6 +159,8 @@ def simplify_objdump(
         row = row.rstrip()
         if ">:" in row or not row:
             continue
+        if "(r2)" in row:
+            row = row.replace("(r2)", "(0)")
         if "R_MIPS_" in row:
             prev = output_lines[-1]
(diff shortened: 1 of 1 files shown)


src/objdump.py
```diff
@@ -159,6 +162,9 @@ def simplify_objdump(
         row = row.rstrip()
         if ">:" in row or not row:
             continue
+        # Fix for ppc sda21 reloc
+        if arch.name == "ppc":
```
simonlindholm, line 166:
```suggestion
        if arch.name == "ppc":
            # With sda21 relocs, the linker transforms `r0` into `r2`/`r13`, and
            # we may encounter this in either pre-transformed or post-transformed
            # versions depending on if the .o file comes from compiler output or
            # from disassembly. Normalize, to make sure both forms are treated as
            # equivalent.
```

simonlindholm (changes requested):

PR #115, opened 2022-05-27
fix reloc support for ppc, and ignore diffs to symbols with catch all symbols

snuffysasa:
Making this a draft pull request, because I still have questions and I assume you will have feedback.

### DIFF
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -241,9 +241,8 @@ def simplify_objdump(
         if ">:" in row or not row:
             continue
 
-        if (
-            arch.re_reloc in row
-        ):  # Process Relocations, modify the previous line and do not add this line to output
+        if arch.re_reloc in row:
+            # Process Relocations, modify the previous line and do not add this line to output
             modified_prev = process_reloc(row, output_lines[-1])
             if modified_prev != -1:
                 output_lines[-1] = modified_prev
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -179,13 +179,13 @@ def process_reloc(reloc_row: str, prev: str) -> Optional[str]:
             pass
         elif "R_PPC_ADDR16_HI" in reloc_row:
             # absolute hi of addr
-            repl = f"{repl}@h"
+            repl = f"{repl}@h"  ## not actually needed by permutor
         elif "R_PPC_ADDR16_HA" in reloc_row:
             # adjusted hi of addr
-            repl = f"{repl}@ha"
+            repl = f"{repl}@ha"  ## not actually needed by permutor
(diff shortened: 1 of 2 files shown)


src/objdump.py
```diff
@@ -32,6 +32,7 @@ class ArchSettings:
     re_reg: Pattern[str]
     re_sprel: Pattern[str]
     re_includes_sp: Pattern[str]
+    re_reloc: str
```
simonlindholm, line 35: the name `re_reloc` indicates that it's a Pattern[str]

src/objdump.py
```diff
@@ -150,6 +155,81 @@ def parse_relocated_line(line: str) -> Tuple[str, str, str]:
     return before, imm, after
 
 
+def process_reloc(reloc_row: str, prev: str):
+    if prev == "<skipped>":
+        return -1
```
simonlindholm, line 160: -1 is a bad sentinel, None please

src/objdump.py
```diff
@@ -150,6 +155,81 @@ def parse_relocated_line(line: str) -> Tuple[str, str, str]:
     return before, imm, after
 
 
+def process_reloc(reloc_row: str, prev: str):
```
simonlindholm, line 158: return type; I think `mypy` will complain as is, please check

src/scorer.py
```diff
@@ -55,6 +53,18 @@ def score(self, cand_o: Optional[str]) -> Tuple[int, str]:
...
 
+        def imm_matches_everything(row: str, arch: ArchSettings) -> bool:
+            if arch.name == "ppc":
+                return any(str in row for str in ["...data", "@"])
+            else:
+                return "(." in row
+
+        def is_valid_symbol(str: str) -> bool:
+            return (
+                re.sub(re.compile(r"[A-Za-z_$][A-Za-z0-9_$]*"), "", str.split("@")[0])
```
simonlindholm, line 64: there are functions in `re` for checking whether a regex matches another string, you don't have to do search-replace

src/objdump.py
snuffysasa: good point, I'll fix

src/objdump.py
snuffysasa: k

src/objdump.py
snuffysasa: ohh, I thought black was auto adding return types... but yea I need to install, mypy. I'll do that and add return type

src/scorer.py
snuffysasa: on it

snuffysasa:
I made a replacement PR on a new branch that I made, closing this one

PR #117, opened 2022-05-30
add feature flag to add minimum score threshold

snuffysasa:
pretty simple, not too much to say.  code is pretty self explanatory

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -55,7 +55,7 @@ class Options:
     abort_exceptions: bool = False
     better_only: bool = False
     best_only: bool = False
-    score_threshold: int = 1
+    score_threshold: Optional[int] = None
     quiet: bool = False
     stop_on_zero: bool = False
     keep_prob: float = DEFAULT_RAND_KEEP_PROB
@@ -648,7 +648,7 @@ def main() -> None:
         "--only-if-below",
         dest="score_threshold",
         type=int,
-        help="The minimum diff score for reporting",
+        help="Only report scores better(lower) than this value",
     )
 
     args = parser.parse_args()
--- a/src/main.py
+++ b/src/main.py
@@ -55,6 +55,7 @@ class Options:
     abort_exceptions: bool = False
     better_only: bool = False
     best_only: bool = False
+    minimum_score: int = 1
     quiet: bool = False
     stop_on_zero: bool = False
     keep_prob: float = DEFAULT_RAND_KEEP_PROB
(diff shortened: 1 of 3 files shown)


simonlindholm (commented):
It's more of a max score than a min score, isn't it? Or since min/max seems a bit ambiguous, perhaps a threshold, rather, or "only if below". Would be neat to fold better-only into this but I don't know how the cmdline flag UX would look for that

snuffysasa:
I would call it a min score, not a max score...  
but --only-if-below could be more clear yea.   

yea it's basically the same thing as --better-only... if the user knows what the base_score is going to be.

but they could run it once to see, then run again with the flag.  Or we could just leave them as separate options.

or it you put the value to 0 it behaves like --better-only ?

src/main.py
```diff
@@ -55,6 +55,7 @@ class Options:
     abort_exceptions: bool = False
     better_only: bool = False
     best_only: bool = False
+    score_threshold: int = 1
```
simonlindholm, line 58: this is an Optional[int], and the default should probably be None

src/permuter.py
```diff
@@ -118,6 +119,7 @@ def __init__(
         self._show_errors = show_errors
         self._best_only = best_only
         self._better_only = better_only
+        self.score_threshold = score_threshold
```
simonlindholm, line 122: make this member private, just like the other ones

src/main.py
```diff
@@ -642,6 +644,12 @@ def main() -> None:
         action="store_true",
         help="Removes all text above the target function in the output source files",
     )
+    parser.add_argument(
+        "--only-if-below",
+        dest="score_threshold",
+        type=int,
+        help="The minimum diff score for reporting",
```
simonlindholm, line 651: this help text is confusing

src/main.py
snuffysasa: thought I had it as optional int,  will update

src/permuter.py
snuffysasa: fixed, yea I had it private, but then when I did the rename, I accidentally changed it

src/main.py
snuffysasa: updated

src/main.py
```diff
@@ -648,7 +648,7 @@ def main() -> None:
         "--only-if-below",
         dest="score_threshold",
         type=int,
-        help="The minimum diff score for reporting",
+        help="Only report scores better(lower) than this value",
```
simonlindholm, line 651:
```suggestion
        help="Only report scores better (lower) than this value",
```

simonlindholm (approved):

PR #119, opened 2022-06-03
PPC reloc fixes

snuffysasa:
Also does various other refactoring and possible fixes outside of ppc too.

This matches the latest design decisions from asm-differ PR # 84

I am closing the previous PR, that this is build off of

### DIFF
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -118,7 +118,7 @@ PPC_SETTINGS: ArchSettings = ArchSettings(
     name="ppc",
     re_includes_sp=re.compile(r"\b(r1)\b"),
     re_comment=re.compile(r"(<.*>|//.*$)"),
-    re_reg=re.compile(r"\$?\b([rf][0-9]+)\b"),
+    re_reg=re.compile(r"(\$?\b([rf][0-9]+)\b|\(.+\))"),
     re_sprel=re.compile(r"(?<=,)(-?[0-9]+|-?0x[0-9a-f]+)\(r1\)"),
     reloc_str="R_PPC_",
     objdump=["powerpc-eabi-objdump", "-dr", "-EB", "-mpowerpc", "-M", "broadway"],
@@ -270,6 +270,10 @@ def simplify_objdump(
             continue
 
         row = re.sub(arch.re_comment, "", row)
+        row = row.rstrip()
+        row = "\t".join(row.split("\t")[2:])  # [20:]
+        if not row:
+            continue
 
         if "\t" in row:
             row_parts = row.split("\t", 1)
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -241,9 +241,8 @@ def simplify_objdump(
         if ">:" in row or not row:
             continue
 
-        if (
-            arch.re_reloc in row
-        ):  # Process Relocations, modify the previous line and do not add this line to output
+        if arch.re_reloc in row:
+            # Process Relocations, modify the previous line and do not add this line to output
             modified_prev = process_reloc(row, output_lines[-1])
             if modified_prev != -1:
                 output_lines[-1] = modified_prev
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -179,13 +179,13 @@ def process_reloc(reloc_row: str, prev: str) -> Optional[str]:
             pass
         elif "R_PPC_ADDR16_HI" in reloc_row:
             # absolute hi of addr
-            repl = f"{repl}@h"
+            repl = f"{repl}@h"  ## not actually needed by permutor
         elif "R_PPC_ADDR16_HA" in reloc_row:
             # adjusted hi of addr
-            repl = f"{repl}@ha"
+            repl = f"{repl}@ha"  ## not actually needed by permutor
         elif "R_PPC_ADDR16_LO" in reloc_row:
             # lo of addr
-            repl = f"{repl}@l"
+            repl = f"{repl}@l"  ## not actually needed by permutor
         elif "R_PPC_ADDR16" in reloc_row:
             # 16-bit absolute addr
             if "+0x7" in repl:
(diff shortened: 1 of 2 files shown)


src/objdump.py
```diff
@@ -24,6 +25,12 @@
 re_int_full = re.compile(r"\b[0-9]+\b")
 
 
+@dataclass
+class Line:
+    row: str
```
simonlindholm, line 30: include the mnemonic here, since we already parse it out and use it in the scorer

src/objdump.py
```diff
@@ -150,10 +162,104 @@ def parse_relocated_line(line: str) -> Tuple[str, str, str]:
...
+        assert any(
+            r in reloc_row for r in ["R_PPC_REL24", "R_PPC_ADDR16", "R_PPC_EMB_SDA21"]
+        ), f"unknown relocation type '{reloc_row}' for line '{prev}'"
+
+        if "R_PPC_REL24" in reloc_row:
+            # function calls
+            pass
+        elif "R_PPC_ADDR16_HI" in reloc_row:
+            # absolute hi of addr
+            repl = f"{repl}@h"  ## not actually needed by permutor
```
simonlindholm, line 214: Permut**e**r, but also, aren't these markers actually used by the scorer?

No need for duplicated comments in all the cases, at least. If we do need a comment it could go at the top of this function and speak in general terms, because I think the same applies to MIPS as for PPC

src/scorer.py
```diff
@@ -33,17 +33,18 @@ def __init__(self, target_o: str, *, stack_differences: bool):
         self.arch = get_arch(target_o)
         self.stack_differences = stack_differences
         _, self.target_seq = self._objdump(target_o)
-        self.differ: difflib.SequenceMatcher[DiffAsmLine] = difflib.SequenceMatcher(
+        self.differ: difflib.SequenceMatcher[str] = difflib.SequenceMatcher(
             autojunk=False
         )
-        self.differ.set_seq2(self.target_seq)
+        self.differ.set_seq2(list(map(lambda x: x.mnemonic, self.target_seq)))
```
simonlindholm, line 39:
```suggestion
        self.differ.set_seq2([line.mnemonic for line in self.target_seq])
```
and same below

src/scorer.py
```diff
@@ -4,19 +4,19 @@
...
-
-from .objdump import objdump, get_arch
+from .objdump import ArchSettings, objdump, get_arch
 
 
 @dataclass(init=False, unsafe_hash=True)
 class DiffAsmLine:
     line: str = field(compare=False)
     mnemonic: str
+    has_symbol: bool
```
simonlindholm, line 14: given the creation of Line in objdump.py and only mnemonics being passed to SequenceMatcher, is DiffAsmLine still needed?

src/objdump.py
```diff
@@ -109,9 +118,10 @@ class ArchSettings:
     name="ppc",
     re_includes_sp=re.compile(r"\b(r1)\b"),
     re_comment=re.compile(r"(<.*>|//.*$)"),
-    re_reg=re.compile(r"\$?\b([rf][0-9]+)\b"),
+    re_reg=re.compile(r"(\$?\b([rf][0-9]+)\b|\(.+\))"),
```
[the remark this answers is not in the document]
snuffysasa: Huh... I thought I just copied over all the re_reg from asm_differ...

I don't know where that `\(.+\))` came from

removing

src/objdump.py
snuffysasa: atleast for PPC it's not needed because we do the function `field_matches_any_symbol` and doesn't care if it has a @ ha suffix

or otherwise it just checks if lhs==rhs.. so we can either add the suffix to both sides or neither side.. both ways they match.

also I have the debug mode now, which prints the asm, so I think its helpful to see the @sda21|ha there.  and also just makes the code more similar to asm-differ 


I am moving the comment at the top of the function, I am not certain about MIPS if its the same or not.  Or maybe I should just remove the comment all together

snuffysasa: yea I am deciding to just remove the comment all together, but let me know if you think one should be there

src/scorer.py
snuffysasa: hmmm! yea we can definitely consolidate here

src/scorer.py
```diff
@@ -55,34 +56,28 @@ def score(self, cand_o: Optional[str]) -> Tuple[int, str]:
...
-                old_idx = old_lo
-                new_idx = new_lo
-            elif old_hi != -1 and new_hi != -1:
-                old_idx = old_hi
-                new_idx = new_hi
-            else:
-                return False
+        def field_matches_any_symbol(field: str, arch: ArchSettings) -> bool:
+            if arch.name == "ppc":
+                if "...data" in field:
```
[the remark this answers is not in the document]
snuffysasa: agreed, I was going to do this and will include in these changes

src/objdump.py
```diff
@@ -89,7 +98,7 @@ class ArchSettings:
     "bgt-",
 }
 
-PPC_BRANCH_LIKELY_INSTRUCTIONS = {}
+PPC_BRANCH_LIKELY_INSTRUCTIONS: Set[str] = set({})
```
simonlindholm, line 101:
```suggestion
PPC_BRANCH_LIKELY_INSTRUCTIONS: Set[str] = set()
```
to avoid constructing an empty dict and passing that to the set constructor

src/objdump.py
```diff
@@ -162,70 +269,50 @@ def simplify_objdump(
...
         if ign_regs:
             row = re.sub(arch.re_reg, "<reg>", row)
 
-        row_parts = row.split(None, 1)
-        if len(row_parts) == 1:
-            row_parts.append("")
-        mnemonic, instr_args = row_parts
         if not stack_differences:
-            if mnemonic == "addiu" and arch.re_includes_sp.search(instr_args):
+            if mnemonic == "addiu" and arch.re_includes_sp.search(args):
```
simonlindholm, line 308: (this will need PPC support too. future work though)

src/scorer.py
```diff
@@ -4,19 +4,7 @@
...
-
-@dataclass(init=False, unsafe_hash=True)
-class DiffAsmLine:
-    line: str = field(compare=False)
-    mnemonic: str
-
-    def __init__(self, line: str) -> None:
-        self.line = line
-        self.mnemonic = line.split(None, 1)[0]
+from .objdump import ArchSettings, Line, objdump, get_arch
```
simonlindholm, line 7:
```suggestion

from .objdump import ArchSettings, Line, objdump, get_arch
```
nit: we leave an empty line between built-in imports and ones from within this module (and third-party libs in between there if any)

src/scorer.py
```diff
@@ -55,34 +40,29 @@ def score(self, cand_o: Optional[str]) -> Tuple[int, str]:
...
+        def field_matches_any_symbol(field: str, arch: ArchSettings) -> bool:
+            if arch.name == "ppc":
+                if "..." in field:
+                    return True
+
+                parts = field.rsplit("@", 1)
+                if len(parts) == 2 and parts[1] in {"l", "h", "ha", "sda21"}:
+                    field = parts[0]
+
+                return re.fullmatch((r"^@\d+$"), field) is not None
```
simonlindholm, line 52:
```suggestion
                return re.fullmatch(r"^@\d+$", field) is not None
```

simonlindholm (approved):

PR #121, opened 2022-06-04
add debug mode

snuffysasa:
add debug mode to latest main branch

### DIFF
--- a/src/compiler.py
+++ b/src/compiler.py
@@ -1,24 +1,37 @@
 from typing import Optional
 import tempfile
 import subprocess
-
+import os
+import shutil
 from .helpers import try_remove
 
 
 class Compiler:
-    def __init__(self, compile_cmd: str, *, show_errors: bool) -> None:
+    def __init__(
+        self, compile_cmd: str, *, show_errors: bool, debug_mode: bool
+    ) -> None:
         self.compile_cmd = compile_cmd
         self.show_errors = show_errors
+        self.debug_mode = debug_mode
 
     def compile(self, source: str, *, show_errors: bool = False) -> Optional[str]:
         """Try to compile a piece of C code. Returns the filename of the resulting .o
         temp file if it succeeds."""
-        show_errors = show_errors or self.show_errors
+        show_errors = show_errors or self.show_errors or self.debug_mode
         with tempfile.NamedTemporaryFile(
             prefix="permuter", suffix=".c", mode="w", delete=False
         ) as f:
             c_name = f.name
             f.write(source)
+
+        if self.debug_mode:
+            debug_filepath = os.getcwd() + "/debug_source_" + c_name[-10:]
+            print(
+                "DEBUG MODE: Saving a full copy of base candidate source to ",
+                debug_filepath,
+            )
+            with open(debug_filepath, "w") as f_copy:
+                f_copy.write(source)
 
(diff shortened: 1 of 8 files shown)


src/compiler.py
```diff
@@ -1,25 +1,38 @@
 from typing import Optional
 import tempfile
 import subprocess
-
+import os
+import shutil
```
simonlindholm, line 5:
```suggestion
import shutil

```

src/scorer.py
```diff
@@ -111,27 +118,81 @@ def diff_delete(line: str) -> None:
...
+                    new_line = cand_seq[i1 + k]
+                    diff_sameline(old_line, new_line)
             if tag == "replace" or tag == "delete":
                 for k in range(i1, i2):
                     diff_insert(cand_seq[k].row)
             if tag == "replace" or tag == "insert":
                 for k in range(j1, j2):
                     diff_delete(self.target_seq[k].row)
 
+        if self.debug_mode:  ## Print simple asm diff
```
simonlindholm, line 136:
```suggestion
        if self.debug_mode:
            # Print simple asm diff
```

src/objdump.py
```diff
@@ -98,7 +98,7 @@ class ArchSettings:
     "bgt-",
 }
 
-PPC_BRANCH_LIKELY_INSTRUCTIONS: Set[str] = set()
+PPC_BRANCH_LIKELY_INSTRUCTIONS: Set[str] = set({})
```
simonlindholm, line 101:
```suggestion
PPC_BRANCH_LIKELY_INSTRUCTIONS: Set[str] = set()
```

src/compiler.py
```diff
@@ -1,25 +1,38 @@
...
-        show_errors = show_errors or self.show_errors
+        show_errors = show_errors or self.show_errors or self.debug_mode
         with tempfile.NamedTemporaryFile(
             prefix="permuter", suffix=".c", mode="w", delete=False
         ) as f:
             c_name = f.name
             f.write(source)
 
+        if self.debug_mode:
+            debug_filepath = os.getcwd() + "/debug_source_" + c_name[-10:]
```
simonlindholm, line 28: `c_name[-10:]` is very ugly and quite possibly unportable. It's just to generate a random string, right? Can you do that in another way?

Also, could use `"."` instead of os.getcwd()

src/main.py
```diff
@@ -650,6 +653,12 @@ def main() -> None:
         type=int,
         help="Only report scores better (lower) than this value",
     )
+    parser.add_argument(
+        "--debug-mode",
```
simonlindholm, line 657:
```suggestion
        "--debug",
```

src/scorer.py
```diff
@@ -6,6 +6,7 @@
 from collections import Counter
 
 from .objdump import ArchSettings, Line, objdump, get_arch
+from colorama import Fore
```
simonlindholm, line 9: I don't think we should add a dependency on colorama just for this. Do like main.py and hard-code color codes

src/compiler.py
snuffysasa: you know I don't think it's even necessary to make a random string here

src/scorer.py
snuffysasa: makes sense

PR #127, opened 2022-06-12
Add arch settings for ARM32

abaresk:
Adds ARM32 support -- works locally using the example function from https://github.com/laqieer/decomp-permuter-arm.


src/objdump.py
```diff
@@ -128,6 +157,26 @@ class ArchSettings:
...
+    #   - VFP/NEON registers: s0..31, d0..31, q0..15, fpscr, fpexc, fpsid
+    # SP should not be in this list.
+    re_reg=re.compile(
+        r"\$?\b([rq][0-9]|[rq]1[0-5]|pc|lr|[ds][12]?[0-9]|[ds]3[01]|fp(scr|exc|sid))\b"
+    ),
+    re_sprel=re.compile(r"sp, #-?(0x[0-9a-fA-F]+|[0-9]+)\b"),
+    reloc_str="R_ARM_",
+    objdump=["arm-none-eabi-objdump", "-drz", "-EL"],
+    branch_instructions=ARM32_BRANCH_INSTRUCTIONS,
+    branch_likely_instructions=ARM32_BRANCH_LIKELY_INSTRUCTIONS,
```
simonlindholm, line 176:
```suggestion
    branch_likely_instructions=set(),
```

simonlindholm (commented):
It could be useful to implement `field_matches_any_symbol` for `arm` to get zero scores for things that match except where the generated code has `.rodata+0x...` and the target asm has some symbol. Same in asm-differ. No big deal though. (PPC cares about it for asm-differ because they use decomp.me a lot and aim for scratches getting marked as "100% matching".)

Also, there's a check
```py
            if mnemonic == "addiu" and arch.re_includes_sp.search(args):
```
that assumes a MIPS mnemonic for add which would be good to generalize to cover ARM. (It aims to not penalize imm differences that are due to stack at "add $out, $sp, <offset>" instructions.)

src/objdump.py
```diff
@@ -189,6 +239,66 @@ def pre_process(mnemonic: str, args: str, next_row: Optional[str]) -> Tuple[str,
...
+            if int(repl.split("+")[1], 16) > 0x70000000:
+                return repl.split("+")[0]
+    elif "R_PPC_EMB_SDA21" in reloc_row:
+        # sda21 relocations; r2/r13 --> 0 swaps are performed in an earlier processing step
+        return f"{repl}@sda21"
+
+
+def process_arm32_reloc(reloc_row: str, prev: str) -> Optional[str]:
+    assert any(
+        r in reloc_row for r in ["R_ARM_THM_CALL", "R_ARM_CALL", "R_ARM_ABS32"]
```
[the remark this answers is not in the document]
abaresk: `R_ARM_ABS32` is used for static and global variables, at least with agbcc and mwccarm.

These are the only relocs I've seen so far. I figure it shouldn't be too hard to expand the list if people encounter others

src/objdump.py
```diff
@@ -201,60 +311,18 @@ def process_reloc(reloc_row: str, prev: str) -> Optional[str]:
...
+    elif "R_PPC_" in reloc_row:
+        new_repl = process_ppc_reloc(reloc_row, prev)
+    elif "R_ARM_" in reloc_row:
+        new_repl = process_arm32_reloc(reloc_row, prev)
+    else:
+        raise Exception(f"unknown relocation type: {reloc_row}")
 
-    raise Exception(f"unknown relocation type: {reloc_row}")
+    if new_repl is not None:
+        repl = new_repl
```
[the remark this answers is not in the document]
abaresk: Done

src/objdump.py
abaresk: Done

src/objdump.py
```diff
@@ -189,6 +239,66 @@ def pre_process(mnemonic: str, args: str, next_row: Optional[str]) -> Tuple[str,
...
+    assert any(
+        r in reloc_row for r in ["R_PPC_REL24", "R_PPC_ADDR16", "R_PPC_EMB_SDA21"]
+    ), f"unknown relocation type '{reloc_row}' for line '{prev}'"
+
+    if "R_PPC_REL24" in reloc_row:
+        # function calls
+        return None
+    elif "R_PPC_ADDR16_HI" in reloc_row:
+        # absolute hi of addr
+        return f"{repl}@h"
```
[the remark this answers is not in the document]
abaresk: Done, thanks for the tip!

src/objdump.py
```diff
@@ -128,6 +157,26 @@ class ArchSettings:
...
+    #   - General purpose registers: r0..13
+    #   - Frame pointer registers: lr (r14), pc (r15)
+    #   - VFP/NEON registers: s0..31, d0..31, q0..15, fpscr, fpexc, fpsid
+    # SP should not be in this list.
+    re_reg=re.compile(
+        r"\$?\b([rq][0-9]|[rq]1[0-5]|pc|lr|[ds][12]?[0-9]|[ds]3[01]|fp(scr|exc|sid))\b"
+    ),
+    re_sprel=re.compile(r"sp, #-?(0x[0-9a-fA-F]+|[0-9]+)\b"),
+    reloc_str="R_ARM_",
+    objdump=["arm-none-eabi-objdump", "-drz", "-EL"],
```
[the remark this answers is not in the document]
abaresk: Removed this, -EL is not required

src/scorer.py
```diff
@@ -59,6 +59,10 @@ def field_matches_any_symbol(field: str, arch: ArchSettings) -> bool:
             if arch.name == "mips":
                 return "." in field
 
+            # Example: ".word  gSymbol+0x4"
```
simonlindholm, line 62: This is interesting if true. This function is meant to handle instruction arguments and treat e.g. `lw a0, %lo(.rodata+0x8)(a0)` the same as `lw a0, %lo(D_12345678)(a0)`, where the .rodata+0x8 is compiler-generated and not something that makes sense to put in the target assembly. The mnemonic itself is currently included in the comparison because we split on `,` without stripping away the mnemonic first; this is a bug that might end up getting fixed by snuffysasa's work, though it's a harmless one. So I'm guessing this works for you, because arm mixes code and data in the .text section and this catches onto the "mnemonic" `.word` and uses that to treat `.word .rodata+0x8` the same as `.word D_12345678`. Is that the case? If so could we do something to make this work even once the bug that makes us include mnemonic as part of the first field gets fixed? (Would it work to change this logic to `return "." in field` just like MIPS?)

abaresk: Yeah I think that's exactly what's happening. You can see it in action with this scratch https://decomp.me/scratch/tadkS.

If the "mnemonic" `.word` gets stripped, `return "." in field` would cover some cases, such as `.word   .text+0x34c` seen here https://decomp.me/scratch/z6Nxz. (This was part of the target assembly, but I'm assuming it can also be generated by the compiler in some cases). But it wouldn't cover the first scratch.

Perhaps, `return "." in field or "+" in field` would work though, although it doesn't quite match the semantics of `field_matches_any_symbol`. Alternatively, in the bug fix, is there a way we could keep certain "mnemonics" like `.word`?

simonlindholm: Curious. In the case of `.word   ov76_021E7094` vs `.word   ov76_021E7098+0x4` I'd be inclined to say the right fix is to change the target asm -- `ov76_021E7098[i].foo` and `ov76_021E7098[i].bar` are part of the same symbol so it makes sense for the asm to refer to them through offsets. And for `.text+0x34c` `return "." in field` should work. Would you be fine with `return "." in field`? The advantage of that is that if there are real symbol differences in `.word` they would affect the score/show up as asm-differ diffs.

Interesting also to see noise like `ComputeDisplayTime-0x4+0x4` in the second diff, looks like there's some low-hanging fruit for improvements. (In cases like `bl      f8` vs `bl      CARDi_WaitPrevCommand-0x8+0x8` I think you could get `CARDi_WaitPrevCommand` on the lhs by marking that as a global symbol.)

abaresk: Apologies, that was a contrived target asm that I wrote up by hand, and I didn't notice that issue. Here's an example from an actual function, which shows why `return "." in field` might not always work: https://decomp.me/scratch/yhy5G.

And I agree, I'm hoping to take a look at making improvements like this after porting changes here to asm-differ. Stuff like`ComputeDisplayTime-0x4+0x4` comes up a lot with mwccarm objects. Also `adds    r0, r0, #1` vs. `adds    r0, #1` seen in the scratch above is a frequent problem. There's trickier issues too, like recognizing jump tables (e.g. https://decomp.me/scratch/Aucj4), but this is all out of scope for this PR I think.

simonlindholm: That scratch seems to have the same property as the previous one, where it would be better to change the target asm.

> There's trickier issues too, like recognizing jump tables (e.g. https://decomp.me/scratch/Aucj4), but this is all out of scope for this PR I think.

For sure. The permuter ignores jump table contents on mips fwiw and it's been fine. For asm-differ it might matter more.

abaresk: Oh I see what you mean. Yeah we should probably switch to symbol offsets in our [codebase](https://github.com/pret/pokeheartgold). I don't know how tractable it will be though in the short-term.

Do you imagine this check will get any more specific in the foreseeable future? Right now, it seems that for other architectures, `field_matches_any_symbol` doesn't care whether it's the correct symbol. If it's going to stay that way, maybe we can do the same for ARM and shoot for a solution that works with both `.text+offset` and `symbol_name`?

Otherwise, I'm happy to go with `return "." in field` for now. I think this would incidentally and temporarily solve the scoring  issue, at least for our codebase. And maybe we can revisit this when mnemonics (and `.word`) get stripped.

simonlindholm: Also it should be said that even if we fix the bug and you haven't transitioned your codebase yet, it's not a big deal: perfect permuter scores will be slightly non-zero but it won't affect effectiveness.

abaresk: Okay, that's good to know - updated this accordingly!

PR #129, opened 2022-06-15
Compiler type specific weights and custom weights

snuffysasa:
Randomization method weights are compiler type specific now and the user can specify custom weight overrides.

Created default weights for compiler types: base, ido, mwcc, and gcc, however the weights are all the same right now

Import.py generates a settings.toml file which stores the compiler type and/or custom weight overrides.

This PR matches our discord discussion


default_weights.toml
```diff
@@ -0,0 +1,119 @@
...
+perm_commutative = 5
+perm_add_sub = 5
+perm_inequalities = 5
+perm_compound_assignment = 5
+perm_remove_ast = 5
+perm_duplicate_assignment = 5
+perm_chain_assignment = 5
+perm_pad_var_decl = 1
+
+[ido]
```
simonlindholm, line 31: These should extend base, don't include the values that are the same.

If you want some non-default weights to start out with, `float_literal` and `sameline` could be reduced (10 -> 3?) for non-IDO

default_weights.toml
```diff
@@ -0,0 +1,119 @@
+[base]
```
simonlindholm, line 1: Needs a comment at the top explaining what this file is for and that you can customize weights in the non-matching dir. Also needs a mention in the README.

import.py
```diff
@@ -644,8 +655,22 @@ def compile_base(compile_script: str, source: str, c_file: str, out_file: str) -
         print("Warning: failed to compile .c file.")
 
 
+def write_custom_weight_comments(compiler_type: str, filename: str) -> None:
+    if os.path.exists("default_weights.toml"):
+        with open("default_weights.toml") as f:
+            all_weights = toml.load(f)
+            compiler_weights = all_weights[compiler_type]
```
simonlindholm, line 662: We'll want to do this in src/ somewhere too, let's use a common helper function. We can assume that `default_weights.toml` always exists and raise an error if not.

import.py
```diff
@@ -644,8 +655,22 @@ def compile_base(compile_script: str, source: str, c_file: str, out_file: str) -
...
+            f.write("# uncomment lines below to customize the weights\n")
+            f.write("# run --help=randomization to see doc\n")
+            f.write("[custom_weights]\n")
+            for randomization_type, weight in compiler_weights.items():
+                f.write("# " + randomization_type + " = " + str(weight) + "\n")
+
+
 def write_to_file(cont: str, filename: str) -> None:
-    with open(filename, "w", encoding="utf-8") as f:
+    with open(filename, "a", encoding="utf-8") as f:
```
simonlindholm, line 673: ehh, let's do string concatenation in the caller instead

import.py
```diff
@@ -717,6 +742,7 @@ def main() -> None:
                 settings = toml.load(f)
             break
 
+    compiler_type = cast(str, settings.get("compiler_type", "base"))
```
simonlindholm, line 745: assert isinstance

src/randomizer.py
```diff
@@ -2007,46 +2044,45 @@ def visit_Decl(self, decl: ca.Decl) -> None:
 
 
 class Randomizer:
-    def __init__(self, rng_seed: int) -> None:
+    def __init__(self, rng_seed: int, dir: str) -> None:
```
simonlindholm, line 2047: randomizer.py should not read from disk: it's better to centralize that responsibility to a single piece of code that reads a nonmatching directory, and furthermore it doesn't work with p@h.

src/randomizer.py
```diff
@@ -83,6 +86,37 @@
 
 T = TypeVar("T")
 
+method_list: List[str] = [
```
simonlindholm, line 89: METHOD_LIST since it's global. Let's make a list of functions instead of strings, to get improved type safety and eliminate the need for using globals(). Decorators could be neat.

import.py
snuffysasa: yea in this case if it doesn't exist, no harm is done. it just doesn't write those commented out lines.   But I can raise an error if you want

default_weights.toml
snuffysasa: done

default_weights.toml
snuffysasa: done

import.py
snuffysasa: done

import.py
snuffysasa: done

import.py
snuffysasa: done

src/randomizer.py
snuffysasa: done

src/randomizer.py
snuffysasa: done

import.py
```diff
@@ -644,6 +656,27 @@ def compile_base(compile_script: str, source: str, c_file: str, out_file: str) -
...
+    all_weights = load_weights_from_file()
+    base_weights = all_weights["base"]
+    compiler_weights = all_weights[compiler_type]
+
+    with open(filename, "w", encoding="utf-8") as f:
+        f.write(f'func_name = "{func_name}"\n')
+        f.write(f'compiler_type = "{compiler_type}"\n\n')
+
+        f.write("# uncomment lines below to customize the weights\n")
+        f.write("# run --help=randomization to see doc\n")
```
simonlindholm, line 672: Let's also implement this if we include it in the comment

import.py
```diff
@@ -644,6 +656,27 @@ def compile_base(compile_script: str, source: str, c_file: str, out_file: str) -
...
+        f.write("[custom_weights]\n")
```
simonlindholm, line 673: (This name bugs me a little: it's kinda implicit that these are custom since we're in a file specifically meant to customize things. Maybe change to something like `[randomization_weights]` or `[perm_weights]` or `[weight_overrides]`? May also be fine to keep though.)

import.py
```diff
@@ -644,6 +656,27 @@ def compile_base(compile_script: str, source: str, c_file: str, out_file: str) -
...
+        for randomization_type, weight in base_weights.items():
+            if randomization_type in compiler_weights:
+                weight = compiler_weights[randomization_type]
+            f.write("# " + randomization_type + " = " + str(weight) + "\n")
```
simonlindholm, line 677:
```suggestion
            f.write(f"# {randomization_type} = {weight}\n")
```

src/randomizer.py
```diff
@@ -2006,47 +2013,79 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+    perm_chain_assignment,
+    perm_pad_var_decl,
+]
+
+
 class Randomizer:
-    def __init__(self, rng_seed: int) -> None:
+    def __init__(
+        self,
+        settings: Optional[Mapping[str, Any]],
```
simonlindholm, line 2051: A little bit skeptical to passing in a whole unparsed dict like this. Maybe reasonable if we want to add further knobs to randomization in the future though, and not just weights. We do want it to be Mapping[str, object] still, and add type-checking accessors, like `json_prop` in net/core.py -- that function could be renamed and moved to helpers.

README.md
```diff
@@ -37,6 +38,8 @@ build_system = "ninja"
 ```
 Then `import.py` should work as expected if `build.ninja` is at the root of the project.
 
+All of the possible randomizations are assigned a weight value that affects the probability/frequency that the randomization is chosen. These weights are specific to the type of compiler because different randomizations are more or less helpful depening on the compiler. A list of the default weights and compiler specific weights are listed in default_weights.toml. Additionally, it is recommended to set the compiler type in the settings.toml file in the input directory, in practice this is done by setting the compiler type in permuter_settings.toml of the project you are matching, or set in the decompme_mappings.toml. You can override the set weight values for a specific function by setting the weights for each pass type individually in the settings.toml file in the input dir. When using import.py it will generate comment lines for customizing that functions wights as reference.
```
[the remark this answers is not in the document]
snuffysasa: "What's decompme_mappings.toml?"

Doesn't exist now ha, but was leaving it there in case we forget when adding the import from decompme URL

src/helpers.py
```diff
@@ -29,3 +30,16 @@ def trim_source(source: str, fn_name: str) -> str:
...
             return source[new_index:]
     return source
+
+
+def load_weights_from_file() -> Mapping[str, Any]:
+    with open("default_weights.toml") as f:
+        return toml.load(f)
+
+
+def load_settings_from_file(dir: str) -> Optional[Mapping[str, Any]]:
```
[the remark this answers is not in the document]
snuffysasa: wait, we can't use assert here because you wanted to support the case where the user does not have a settings.toml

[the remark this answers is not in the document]
snuffysasa: that's why I made it optional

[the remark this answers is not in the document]
snuffysasa: nervermind, makes sense now after reading below comments

src/helpers.py
```diff
@@ -29,3 +30,16 @@ def trim_source(source: str, fn_name: str) -> str:
...
+    if os.path.exists(os.path.join(dir, "settings.toml")):
+        with open(os.path.join(dir, "settings.toml")) as f:
+            return toml.load(f)
+    else:
+        return None
```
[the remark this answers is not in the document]
snuffysasa: done

src/permuter.py
```diff
@@ -77,6 +78,8 @@ def __init__(
         scorer: Scorer,
         source_file: str,
         source: str,
+        settings: Optional[Mapping[str, Any]],
+        weights: Mapping[str, Any],
```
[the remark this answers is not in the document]
snuffysasa: okay

src/randomizer.py
```diff
@@ -2006,47 +2013,79 @@ def visit_Decl(self, decl: ca.Decl) -> None:
     ast_util.insert_decl(fn, var, type, random)
 
 
+randomization_passes: List[Callable] = [
```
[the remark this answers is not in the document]
snuffysasa: done

import.py
snuffysasa: I like weight_overrides

src/randomizer.py
snuffysasa: I just finished refactoring it so the weight overrides are applied in main, and settings is no longer passed to randomizer

just randomization_weights: Mapping[str, float]

src/helpers.py
```diff
@@ -32,14 +32,22 @@ def trim_source(source: str, fn_name: str) -> str:
     return source
 
 
-def load_weights_from_file() -> Mapping[str, Any]:
+def load_weights_from_file(compiler_type: str) -> Mapping[str, float]:
```
simonlindholm, line 35: `load_weights_from_file` -> `get_default_randomization_weights`, to make usage of the term "randomization weights" more consistent, and because "from file" is a bit of an implementation detail

src/randomizer.py
```diff
@@ -2048,36 +2050,18 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
-                weight = custom_weights[func_name]
-            elif func_name in compiler_weights:
-                weight = compiler_weights[func_name]
-            else:
-                weight = base_weights[func_name]
-
-            return (method, weight)
-
-        self.methods = list(map(create_method_weight_pair, randomization_passes))
+        self.methods = list(map(create_method_weight_pair, RANDOMIZATION_PASSES))
```
simonlindholm, line 2064:
```suggestion
        self.methods = [(method, randomization_weights[method.__name__]) for method in RANDOMIZATION_PASSES]
```
now that the inner function is a one-liner

src/helpers.py
```diff
@@ -32,14 +32,22 @@ def trim_source(source: str, fn_name: str) -> str:
...
+    weights: Dict[str, float] = {}
     with open("default_weights.toml") as f:
-        return toml.load(f)
+        all_weights = toml.load(f)
+        base_weights = all_weights["base"]
+        compiler_weights = all_weights[compiler_type]
+        for randomization_type, weight in base_weights.items():
+            if randomization_type in compiler_weights:
+                weight = compiler_weights[randomization_type]
+            weights[randomization_type] = weight
```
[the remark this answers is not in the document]
snuffysasa: I did switch to `key` and thats much better

src/helpers.py
```diff
@@ -32,14 +32,22 @@ def trim_source(source: str, fn_name: str) -> str:
     return source
 
 
-def load_weights_from_file() -> Mapping[str, Any]:
+def load_weights_from_file(compiler_type: str) -> Mapping[str, float]:
+    weights: Dict[str, float] = {}
     with open("default_weights.toml") as f:
-        return toml.load(f)
+        all_weights = toml.load(f)
```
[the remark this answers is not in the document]
snuffysasa: same as I messaged you on discord for the other spot in Main... I can't type this as object without a case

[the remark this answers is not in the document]
snuffysasa: "Any means we may have incorrect types here that can inflict the rest of the program and cause errors at a distance."

Well this isn't really true in this case right? and in the other case too.
Because this function returns type Mapping[str, float]... so we know we are getting the correct types there... so wouldn't cause errors at a distance.

I only used Any in locations where the data gets immediately pulled out and type checked and then the Any object is discarded.

src/helpers.py
snuffysasa: agree that is a better name

src/helpers.py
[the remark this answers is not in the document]
simonlindholm: `compiler_weights.get(randomization_type, weight)` does work: the second parameter provides a fallback for when the key does not exist in the map.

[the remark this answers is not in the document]
snuffysasa: I see, updating to that

README.md
```diff
@@ -37,6 +38,13 @@ build_system = "ninja"
 ```
 Then `import.py` should work as expected if `build.ninja` is at the root of the project.
 
+All of the possible randomizations are assigned a weight value that affects the probability/frequency that the randomization is chosen.
+These weights are specific to the type of compiler because different randomizations are more or less helpful depending on the compiler.
+A list of the default weights and compiler specific weights are listed in default_weights.toml.
+Additionally, it is recommended to set the compiler type in the settings.toml file in the input directory, in practice this is done by setting the compiler type in permuter_settings.toml of the project you are matching.
+You can override the set weight values for a specific function by setting the weights for each pass type individually in the settings.toml file in the input dir.
+When using import.py it will generate comment lines for customizing that functions weights as reference.
```
simonlindholm, line 46:
```suggestion
All of the possible randomizations are assigned a weight value that affects the frequency with which the randomization is chosen.
The default set of weights is specified in `default_weights.toml` and vary based on the targeted compiler.
These weights can be overridden by modifying `settings.toml` in the input directory.
```

src/randomizer.py
```diff
@@ -1,5 +1,8 @@
 import bisect
 import copy
+import os
+import sys
+import toml
```
simonlindholm, line 5:
```suggestion
import sys
```

simonlindholm (commented):
Starting to look good!

snuffysasa:
Just updated with these suggested tweaks

import.py
```diff
@@ -743,6 +743,13 @@ def main() -> None:
 
     assert isinstance(compiler_type, str)
 
+    if compiler_type is "base":
+        print(
+            """Warning: Compiler type is missing from this project's permuter settings. 
+        Defaulting to base compiler randomization settings. For best permtuation results, 
+        please set 'compiler_type' in this project's permuter_settings.toml  """
```
simonlindholm, line 750:
```suggestion
            "Warning: Compiler type is missing from this project's permuter settings.\n"
            "Defaulting to base compiler randomization settings. For best permutation results,\n"
            "please set 'compiler_type' in this project's permuter_settings.toml."
```

import.py
```diff
@@ -743,6 +743,13 @@ def main() -> None:
 
     assert isinstance(compiler_type, str)
 
+    if compiler_type is "base":
```
simonlindholm, line 746: `"base"` is fine imo, it's just if the key is unset that we should warn. Also, never use `is` comparisons to compare literals, it's random whether it will work or not.

Maybe we should print `f"Compiler type: {compiler_type}"` if it's set, btw. So like:
```py
compiler_type = settings.get("compiler_type")
if compiler_type is not None:
    assert isinstance(compiler_type, str)
    print(f"Compiler type: {compiler_type}")
else:
    compiler_type = "base"
    print(...)
```

simonlindholm (commented):
Please address

simonlindholm:
> Please address

Oops, incomplete sentence, meant to include "... the USAGE.md thing"

import.py
snuffysasa: okay done

example_settings.toml
```diff
@@ -0,0 +1,34 @@
+func_name = "func_800123456"
+compiler_type = "ido"
+
+# uncomment lines below to customize the weights
+# see README.md
```
simonlindholm, line 5: imo for the example it's better to just include a single override here -- we don't want to maintain the commented-out weights

example_settings.toml
```diff
@@ -0,0 +1,7 @@
+func_name = "func_800123456"
+compiler_type = "ido" # examples: base, ido, mwcc, gcc
+
+# uncomment lines below to customize the weights
+# see README.md
```
simonlindholm, line 5:
```suggestion
```

example_settings.toml
```diff
@@ -0,0 +1,7 @@
...
+[weight_overrides]
+# perm_temp_for_expr = 100
```
simonlindholm, line 7:
```suggestion
perm_temp_for_expr = 100
```

PR #130, opened 2022-06-17
fix randomizer to save return types of in-function variable function call declarations

snuffysasa:
fix randomizer to save return types of in-function variable function call declarations


simonlindholm:
Example? I can't parse the issue title.

PR #134, opened 2022-06-20
importDecompme

snuffysasa:
Reopening this as WIP PR,

I know updates still need to be made to support different compiler types / setups.  Right now it works for mwcc though.

### DIFF
--- a/example_decompme_mappings.toml
+++ b/example_decompme_mappings.toml
@@ -1,0 +1,8 @@
+# Decompme Mappings
+
+[decompme_compiler_identifier]
+compiler_type = "mwcc" #compiler type that matches the options in default_weights.toml
+PATH = "~/Personal/melee/tools/mwcc_compiler/1.2.5"  ## path to the directory where the compiler command should be ran from
+EXE_NAME = "mwcceppc.exe"  ## Compiler exe name
+USE_WINE = true ## Set this to true if you need to use wine for this compiler (using a windows compiler on linux)
+
--- a/import.py
+++ b/import.py
@@ -4,6 +4,8 @@ import argparse
 from collections import defaultdict
 import json
 import os
+import stat
+import traceback
 import platform
 import re
 import shlex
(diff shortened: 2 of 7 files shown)


r-burns:
What needs to be done to get this to a mergeable state? I've been finding this really useful and would be happy to help out.

example_decompme_mappings.toml
```diff
@@ -0,0 +1,8 @@
+# Decompme Mappings
+
+[decompme_compiler_identifier]
```
ethteck, line 3: I think this settings section could/should just go into permuter_settings.toml unless you think there's a good reason to keep it separate

import.py
```diff
@@ -667,10 +674,146 @@ def write_to_file(cont: str, filename: str) -> None:
...
+        print(traceback.format_exc())
+        print(
+            "Exception occured while creating a compile.sh for the permuter from decompme_mappings.toml and scratch settings"
+        )
+        sys.exit(1)
+
+
 def main() -> None:
+
+    if len(sys.argv) > 1 and sys.argv[1].startswith(DECOMPME_BASE):
```
ethteck, line 805: I'm assuming there's a way to do this that's more argparse-y instead of manually checking argv - that might be better here

ethteck (commented):
Haven't tested this yet, but in the interest of keeping things moving, I perused this and left a couple comments. Hopefully we can work to get it merged :)

simonlindholm:
I created a branch https://github.com/simonlindholm/decomp-permuter/tree/importDecompme which is an updated version of this PR (though it's untested). It lost some steam once I realized it wasn't clear how to handle macro preservation -- normally the permuter can get information about that from permuter_settings.toml but decomp.me doesn't have that.

import.py
simonlindholm: Not really afaik, we need to do different parsing depending on whether something that looks like a URL is passed

example_decompme_mappings.toml
simonlindholm: permuter_settings.toml only exists in repos, but we don't have a repo at the point we do `./import.py https://decomp.me/...`.

PR #139, opened 2022-06-26
tweak creating the replace cand set to have more possible options

snuffysasa:
tweak creating the replace cand set to have more possible options

### DIFF
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -712,11 +712,11 @@ def perm_temp_for_expr(
     else:
         replace_subexprs(fn.body, find_duplicates)
 
+    assert orig_expr in replace_cands
     replace_cand_set: Set[Expression] = set()
     if random_bool(random, PROB_TEMP_REPLACE_ALL):
         replace_cand_set.update(replace_cands)
     elif random_bool(random, PROB_TEMP_REPLACE_MOST):
-        assert orig_expr in replace_cands
         index = replace_cands.index(orig_expr)
         if random_bool(random, 0.5):
             replace_cand_set.update(replace_cands[: index + 1])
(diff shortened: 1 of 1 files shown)


src/randomizer.py
```diff
@@ -712,20 +712,22 @@ def find_duplicates(e: Expression) -> None:
     else:
         replace_subexprs(fn.body, find_duplicates)
 
-    assert orig_expr in replace_cands
```
simonlindholm, line 715: worth keeping this assertion I think

src/randomizer.py
```diff
@@ -712,20 +712,22 @@ def find_duplicates(e: Expression) -> None:
...
+        if random_bool(random, 0.5):
+            replace_cand_set.update(replace_cands[: index + 1])
+        else:
+            replace_cand_set.update(replace_cands[index:])
+    else:
+        replace_cand_set.add(orig_expr)
+
+    for cand in replace_cands:
+        if random_bool(random, 0.5):
+            replace_cand_set.add(cand)
```
simonlindholm, line 730:
```suggestion
    if random_bool(random, 0.5):
        for cand in replace_cands:
            if random_bool(random, 0.5):
                replace_cand_set.add(cand)
```
to prioritize more natural replacement ranges

simonlindholm (approved):

src/randomizer.py
snuffysasa: I had this moved to a different spot, which made sense / was necessary for a different PR I was working on.

But I'll move it back for now, and can discuss moving it for the other PR later

src/randomizer.py
snuffysasa: yea this makes sense

PR #140, opened 2022-06-26
expand perm_chain_assignment, adding a chance to create the chain assignment with a non consecutive statement

snuffysasa:
expand perm_chain_assignment, adding a chance to create the chain assignment with a non consecutive statement

### DIFF
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -1941,21 +1941,20 @@ def perm_duplicate_assignment(
 def perm_chain_assignment(
     fn: ca.FuncDef, ast: ca.FileAST, indices: Indices, region: Region, random: Random
 ) -> None:
-    """Combine two consecutive assignments into one chain assignment."""
+    """Combine two assignments into one chain assignment."""
     cands: List[Tuple[int, int, Block]] = []
 
     def rec(block: Block) -> None:
         statements = ast_util.get_block_stmts(block, False)
 
-        # Chance to create the chain assignment using a non conesecutive statement
-        if random_bool(random, 0.5):
-            shift_range = 1
-        else:
-            shift_range = random.randint(1, 5)
-
-        for shift in range(1, shift_range + 1):
-            zipped_stmts = zip(statements, statements[shift:])
-            for i, (stmt, next_stmt) in enumerate(zipped_stmts):
+        for i, stmt in enumerate(statements):
+            if (
+                not isinstance(stmt, ca.Assignment)
+                or not region.contains_node(stmt)
+                or ast_util.is_effectful(stmt.rvalue)
+            ):
+                continue
+            for j, next_stmt in enumerate(statements[i + 1 : i + 5], i + 1):
                 if (
                     region.contains_node(stmt)
                     and region.contains_node(next_stmt)
--- a/src/randomizer.py
+++ b/src/randomizer.py
@@ -1968,11 +1968,11 @@ def perm_chain_assignment(
     rec(fn.body)
 
     ensure(cands)
-    chosen_assignment_idx, shift, block = random.choice(cands)
+    chosen_assignment_idx, next_stmt_idx, block = random.choice(cands)
 
     statements = ast_util.get_block_stmts(block, True)
     stmt = statements[chosen_assignment_idx]
-    next_stmt = statements[chosen_assignment_idx + shift]
+    next_stmt = statements[next_stmt_idx]
 
     assert isinstance(stmt, ca.Assignment)
     assert isinstance(next_stmt, ca.Assignment)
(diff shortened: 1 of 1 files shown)


src/randomizer.py
```diff
@@ -1942,35 +1942,44 @@ def perm_chain_assignment(
     fn: ca.FuncDef, ast: ca.FileAST, indices: Indices, region: Region, random: Random
 ) -> None:
     """Combine two consecutive assignments into one chain assignment."""
```
simonlindholm, line 1944: Update the doc comment

src/randomizer.py
```diff
@@ -1942,35 +1942,44 @@ def perm_chain_assignment(
...
-                and isinstance(next_stmt, ca.Assignment)
-                and (
-                    ast_util.equal_ast(stmt.lvalue, next_stmt.rvalue)
-                    or ast_util.equal_ast(stmt.rvalue, next_stmt.rvalue)
-                )
-                and not ast_util.is_effectful(next_stmt.rvalue)
-            ):
-                cands.append((i, block))
+
+        # Chance to create the chain assignment using a non conesecutive statement
```
simonlindholm, line 1950: I think we could set shift_range to always be 5 tbh, it doesn't hurt

src/randomizer.py
```diff
@@ -1942,35 +1942,44 @@ def perm_chain_assignment(
...
+                    and region.contains_node(next_stmt)
+                    and isinstance(stmt, ca.Assignment)
+                    and isinstance(next_stmt, ca.Assignment)
+                    and (
+                        ast_util.equal_ast(stmt.lvalue, next_stmt.rvalue)
+                        or ast_util.equal_ast(stmt.rvalue, next_stmt.rvalue)
+                    )
+                    and not ast_util.is_effectful(next_stmt.rvalue)
+                ):
+                    cands.append((i, shift, block))
```
simonlindholm, line 1970: This loops reads a bit funny now. I would suggest rewriting it as something like
```py
for i, stmt in enumerate(statements):
    if not isinstance(stmt, ca.Assignment) or not region.contains_node(stmt) or ast_util.is_effectful(stmt.rvalue):
        continue
    for j, next_stmt in enumerate(statements[i+1:i+5], i+1):
        if ...:
            cands.append((i, j, block))

**Comment by snuffysasa** [comment]
Updated with these suggestions

File: src/randomizer.py [hunk 4]
```diff
@@ -1941,36 +1941,44 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
...
+                or not region.contains_node(stmt)
+                or ast_util.is_effectful(stmt.rvalue)
             ):
-                cands.append((i, block))
+                continue
+            for j, next_stmt in enumerate(statements[i + 1 : i + 5], i + 1):
+                if (
+                    region.contains_node(stmt)
+                    and region.contains_node(next_stmt)
+                    and isinstance(stmt, ca.Assignment)
```
simonlindholm, line 1961:
```suggestion
                    region.contains_node(next_stmt)
```

src/randomizer.py
```diff
@@ -1941,36 +1941,44 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
...
-                and isinstance(stmt, ca.Assignment)
-                and isinstance(next_stmt, ca.Assignment)
-                and (
-                    ast_util.equal_ast(stmt.lvalue, next_stmt.rvalue)
-                    or ast_util.equal_ast(stmt.rvalue, next_stmt.rvalue)
-                )
-                and not ast_util.is_effectful(next_stmt.rvalue)
+                not isinstance(stmt, ca.Assignment)
+                or not region.contains_node(stmt)
+                or ast_util.is_effectful(stmt.rvalue)
```
simonlindholm, line 1954:
```suggestion
```
(I mistakenly thought this could replace the `is_effectful` check for `next_stmt.rvalue` when I made my suggestion)

src/randomizer.py
```diff
@@ -1941,36 +1941,44 @@ def visit_Assignment(self, node: ca.Assignment) -> None:
...
+                    and region.contains_node(next_stmt)
+                    and isinstance(stmt, ca.Assignment)
+                    and isinstance(next_stmt, ca.Assignment)
+                    and (
+                        ast_util.equal_ast(stmt.lvalue, next_stmt.rvalue)
+                        or ast_util.equal_ast(stmt.rvalue, next_stmt.rvalue)
+                    )
+                    and not ast_util.is_effectful(next_stmt.rvalue)
+                ):
+                    cands.append((i, shift, block))
```
simonlindholm, line 1969: `shift` isn't defined. Let's do `j` and adjust the logic below

snuffysasa: in this case won't j still represent a shift value?

an offset from i, rather than an absolute index,  therefore the logic below should stay the same?

I definitely could be confused here though

snuffysasa:
I updated this PR with these changes you suggested.

But I got a little confused on these last requests, and wasn't sure if I did/understood what you were asking for.   

Can you please carefully check the latest version over once more?

snuffysasa:
hmm, looking at this code after I pushed... it seems like we actually want to use:
`cands.append((i, j + 1, block))`
?

snuffysasa:
nevermind, now I understand that enumerate with starting index.  and see that j was an absolute index. 
and understand what you meant by updating the logic below.
I have updated the PR again

PR #141, opened 2022-06-28
perm_long_chain_assignment

snuffysasa:
Create a long chain assignment out of 3 or more consecutive assignments with equal rvalue

### DIFF
--- a/default_weights.toml
+++ b/default_weights.toml
@@ -29,7 +29,7 @@ perm_compound_assignment = 5
 perm_remove_ast = 5
 perm_duplicate_assignment = 5
 perm_chain_assignment = 5
-perm_long_chain_assignment = 1
+perm_long_chain_assignment = 3
 perm_pad_var_decl = 1
 
 [ido]
--- a/default_weights.toml
+++ b/default_weights.toml
@@ -29,6 +29,7 @@ perm_compound_assignment = 5
 perm_remove_ast = 5
 perm_duplicate_assignment = 5
 perm_chain_assignment = 5
+perm_long_chain_assignment = 3
 perm_pad_var_decl = 1
 
 [ido]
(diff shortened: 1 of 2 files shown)


snuffysasa:
Okay I know the nested while loops in the rec(block) function is quite C-style and not very pythonic.

maybe you can help me out?

src/randomizer.py
```diff
@@ -1992,6 +1992,65 @@ def rec(block: Block) -> None:
...
+    statements = ast_util.get_block_stmts(block, True)
+
+    # Merge all statements into long chain assignment at start_idx
+    stmt = statements[start_idx]
+    assert isinstance(stmt, ca.Assignment)
+    for i in range(start_idx + 1, end_idx + 1):
+        additional_stmt = statements[i]
+        assert isinstance(additional_stmt, ca.Assignment)
+        new_lvalue = additional_stmt.lvalue
+        stmt.rvalue = ca.Assignment("=", stmt.lvalue, stmt.rvalue)  # copy of stmt
```
simonlindholm, line 2046:
```suggestion
        stmt.rvalue = ca.Assignment("=", stmt.lvalue, copy.deepcopy(stmt.rvalue))
```

simonlindholm (commented):
Looks good, I can see why you want this separated from perm_chain_assignment (assignments being consecutive, no logic for `a = b; c = a`). I think the nested whiles look very sensible

default_weights.toml
```diff
@@ -29,13 +29,15 @@ perm_compound_assignment = 5
 perm_remove_ast = 5
 perm_duplicate_assignment = 5
 perm_chain_assignment = 5
+perm_long_chain_assignment = 1
```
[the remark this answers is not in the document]
simonlindholm: That's fair! Maybe it should be lower for mwcc too then...?

src/randomizer.py
snuffysasa: wait, can't I just make this whole thing:   
`stmt.rvalue = copy.deepcopy(stmt)`

PR #142, opened 2022-07-01
Perm inline get structmember

snuffysasa:
WIP - sharing what I got so far

### DIFF
--- a/default_weights.toml
+++ b/default_weights.toml
@@ -31,6 +31,7 @@ perm_duplicate_assignment = 5
 perm_chain_assignment = 5
 perm_long_chain_assignment = 3
 perm_pad_var_decl = 1
+perm_inline_get_structmember = 1
 
 [ido]
 perm_float_literal = 10
--- a/default_weights.toml
+++ b/default_weights.toml
@@ -36,6 +36,7 @@ perm_inline_get_structmember = 1
 [ido]
 perm_float_literal = 10
 perm_sameline = 10
+perm_inline_get_structmember = 0
 
 [mwcc]
 perm_chain_assignment = 10
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -236,21 +236,21 @@ def get_block_stmts(block: Block, force: bool) -> List[Statement]:
 
 def make_decl(
     name: str,
-    type: Union[SimpleType, ca.FuncDecl],
-    quals: List[str] = [],
-    align: List[ca.Alignas] = [],
-    storage: List[str] = [],
-    funcspec: List[str] = [],
+    type: "ca.Type",
+    quals: Optional[List[str]] = None,
+    align: Optional[List[ca.Alignas]] = None,
+    storage: Optional[List[str]] = None,
+    funcspec: Optional[List[str]] = None,
     init: Optional[Union[Expression, ca.InitList]] = None,
     bitsize: Optional[Expression] = None,
 ) -> ca.Decl:
     type = copy.deepcopy(type)
     decl = ca.Decl(
         name=name,
-        quals=quals,
-        align=align,
-        storage=storage,
-        funcspec=funcspec,
+        quals=quals or [],
+        align=align or [],
+        storage=storage or [],
+        funcspec=funcspec or [],
         type=type,
         init=init,
         bitsize=bitsize,
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -199,7 +199,7 @@ def is_lvalue(expr: Expression) -> bool:
     return False
 
 
-def find_parent_addressOP(block: Block, target_node: ca.Node) -> Optional[ca.UnaryOp]:
+def find_parent_address_op(block: Block, target_node: ca.Node) -> Optional[ca.UnaryOp]:
     ret: Optional[ca.UnaryOp] = None
 
     class Visitor(ca.NodeVisitor):
@@ -207,20 +207,6 @@ def find_parent_addressOP(block: Block, target_node: ca.Node) -> Optional[ca.Una
             nonlocal ret
             if node.op is "&" and node.expr is target_node:
                 ret = node
-            self.generic_visit(node)
-
-    Visitor().visit(block)
-    return ret
-
-
-def is_assignment_lvalue(block: Block, target_node: ca.Node) -> bool:
-    ret = False
-
-    class Visitor(ca.NodeVisitor):
-        def visit_Assignment(self, node: ca.Assignment) -> None:
-            nonlocal ret
-            if node.lvalue is target_node:
-                ret = True
             self.generic_visit(node)
 
     Visitor().visit(block)
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -237,6 +237,7 @@ def get_block_stmts(block: Block, force: bool) -> List[Statement]:
 def make_decl(
     name: str,
     type: "ca.Type",
+    *,
     quals: Optional[List[str]] = None,
     align: Optional[List[ca.Alignas]] = None,
     storage: Optional[List[str]] = None,
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -199,20 +199,6 @@ def is_lvalue(expr: Expression) -> bool:
     return False
 
 
-def find_parent_address_op(block: Block, target_node: ca.Node) -> Optional[ca.UnaryOp]:
-    ret: Optional[ca.UnaryOp] = None
-
-    class Visitor(ca.NodeVisitor):
-        def visit_UnaryOp(self, node: ca.UnaryOp) -> None:
-            nonlocal ret
-            if node.op is "&" and node.expr is target_node:
-                ret = node
-            self.generic_visit(node)
-
-    Visitor().visit(block)
-    return ret
-
-
 def is_effectful(expr: Expression) -> bool:
     found = False
 
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -199,6 +199,34 @@ def is_lvalue(expr: Expression) -> bool:
     return False
 
 
+def find_parent_addressOP(block: Block, target_node: ca.Node) -> Optional[ca.UnaryOp]:
+    ret: Optional[ca.UnaryOp] = None
+
+    class Visitor(ca.NodeVisitor):
+        def visit_UnaryOp(self, node: ca.UnaryOp) -> None:
+            nonlocal ret
+            if node.op is "&" and node.expr is target_node:
+                ret = node
+            self.generic_visit(node)
+
+    Visitor().visit(block)
+    return ret
+
+
+def is_assignment_lvalue(block: Block, target_node: ca.Node) -> bool:
+    ret = False
+
+    class Visitor(ca.NodeVisitor):
+        def visit_Assignment(self, node: ca.Assignment) -> None:
+            nonlocal ret
+            if node.lvalue is target_node:
+                ret = True
+            self.generic_visit(node)
+
+    Visitor().visit(block)
+    return ret
+
+
 def is_effectful(expr: Expression) -> bool:
     found = False
 
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -234,21 +234,35 @@ def get_block_stmts(block: Block, force: bool) -> List[Statement]:
     return ret
 
 
+def make_decl(
+    name: str,
+    type: Union[SimpleType, ca.FuncDecl],
+    quals: List[str] = [],
+    align: List[ca.Alignas] = [],
+    storage: List[str] = [],
+    funcspec: List[str] = [],
+    init: Optional[Union[Expression, ca.InitList]] = None,
+    bitsize: Optional[Expression] = None,
+) -> ca.Decl:
+    type = copy.deepcopy(type)
+    decl = ca.Decl(
+        name=name,
+        quals=quals,
+        align=align,
+        storage=storage,
+        funcspec=funcspec,
+        type=type,
+        init=init,
+        bitsize=bitsize,
+    )
+    set_decl_name(decl)
+    return decl
+
+
 def insert_decl(
     fn: ca.FuncDef, var: str, type: SimpleType, random: Optional[Random] = None
 ) -> None:
-    type = copy.deepcopy(type)
-    decl = ca.Decl(
-        name=var,
-        quals=[],
-        align=[],
-        storage=[],
-        funcspec=[],
-        type=type,
-        init=None,
-        bitsize=None,
-    )
-    set_decl_name(decl)
+    decl = make_decl(var, type)
     assert fn.body.block_items, "Non-empty function"
     for index, stmt in enumerate(fn.body.block_items):
         if not isinstance(stmt, ca.Decl):
--- a/src/candidate.py
+++ b/src/candidate.py
@@ -35,7 +35,7 @@ class Candidate:
 
     ast: ca.FileAST
 
-    fn: ca.FuncDef
+    fn_name: str
     rng_seed: int
     randomizer: Randomizer
     score_value: Optional[int] = field(init=False, default=None)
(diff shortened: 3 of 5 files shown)


src/randomizer.py
```diff
@@ -2080,6 +2080,102 @@ def visit_Decl(self, decl: ca.Decl) -> None:
     ast_util.insert_decl(fn, var, type, random)
 
 
+def perm_inline_get_structmember(
+    fn: ca.FuncDef, ast: ca.FileAST, indices: Indices, region: Region, random: Random
+) -> None:
+    """Creates an inline for accessing a struct member."""
```
simonlindholm, line 2086:
```suggestion
    """Creates an inline function for accessing a struct member."""
```

src/randomizer.py
```diff
@@ -2080,6 +2080,102 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+
+    for item in ast.ext:
+        if (
+            isinstance(item, ca.Decl)
+            and isinstance(item.type, ca.FuncDecl)
+            and item.name
+        ):
+            ext_fns.append(item.name)
+        if isinstance(item, ca.FuncDef) and item.decl.name:
+            ext_fns.append(item.decl.name)
```
simonlindholm, line 2101: maybe split this out into a helper method for "create non-colliding name"?

src/randomizer.py
```diff
@@ -2080,6 +2080,102 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+
+    typemap = build_typemap(ast)
+
+    ext_fns: List[str] = []
+    cands: List[ca.StructRef] = []
+
+    for item in ast.ext:
+        if (
+            isinstance(item, ca.Decl)
+            and isinstance(item.type, ca.FuncDecl)
```
simonlindholm, line 2096:
```suggestion
```
and rename ext_fns -> used_names and make it a set

src/randomizer.py
```diff
@@ -2080,6 +2080,102 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+        fn.body, cand, ca.FuncCall(ca.ID(new_inline_fn_name), ca.ExprList([cand.name]))
+    )
+
+    assert isinstance(arg_type.type, ca.TypeDecl)
+    assert arg_type.type.declname
+
+    if isinstance(member_type, ca.PtrDecl):
+        assert isinstance(member_type.type, ca.TypeDecl)
+        assert isinstance(member_type.type.type, ca.IdentifierType)
+        member_type_str = member_type.type.type.names[0] + "*"
```
simonlindholm, line 2134: don't create pointers like this, it will break further analysis performed on the AST. (It's also vaguely preferable to use `member_type` in case it's a typedefed pointer)

src/randomizer.py
```diff
@@ -2080,6 +2080,102 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+
+    # Replace the StructRef with a FuncCall to the new inline yet to be created
+    replace_node(
+        fn.body, cand, ca.FuncCall(ca.ID(new_inline_fn_name), ca.ExprList([cand.name]))
+    )
+
+    assert isinstance(arg_type.type, ca.TypeDecl)
+    assert arg_type.type.declname
+
+    if isinstance(member_type, ca.PtrDecl):
```
simonlindholm, line 2131: `resolve_typedefs` before checking if you have a pointer

src/randomizer.py
```diff
@@ -2080,6 +2080,102 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+        assert isinstance(member_type.type, ca.TypeDecl)
+        assert isinstance(member_type.type.type, ca.IdentifierType)
+        member_type_str = member_type.type.type.names[0] + "*"
+    elif isinstance(member_type, ca.TypeDecl):
```
simonlindholm, line 2135: passing a whole struct to an inline function feels unlikely... should we pass &x instead, or bail out?

src/candidate.py
```diff
@@ -70,15 +70,16 @@ def from_source(
         fn_copy = copy.deepcopy(orig_fn)
         ast.ext[fn_index] = fn_copy
         apply_ast_perms(fn_copy, eval_state)
+        assert fn_copy.decl.name
         return Candidate(
             ast=ast,
-            fn_index=fn_index,
+            fn_name=fn_copy.decl.name,
```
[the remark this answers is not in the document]
snuffysasa: ah didn't see fn_name was already there

src/randomizer.py
snuffysasa: done

src/randomizer.py
snuffysasa: check

src/randomizer.py
snuffysasa: okay, I actually originally did not make the pointer like this... but then I was not able to get mypy to behave... mypy forced me to use the `ca.IdentifierType(names=["some_string_here"])`  and if I wanted it to be a pointer I had to manually add the *

I am able to make it cleaner using member_type, but couldn't get mypy to accept it

src/randomizer.py
snuffysasa: I'm confused, I already have a SimpleType.. using resolve_typedefs which just make the type more ambiguous

src/randomizer.py
snuffysasa: This isn't setup to pass a whole struct to an inline function, this is for the return value, which likely would not be a struct.

snuffysasa:
```
inline s32 get_x8_spawnNum(Fighter *fighter2)
{
  return fighter2->x8_spawnNum;
}
```

example member is the spawnNum, in this case an s32

src/randomizer.py
snuffysasa: the current code is working with typedefed pointers

snuffysasa: update -  I got the mypy stuff resolved and made it work using member_type and not manually creating the type string

src/randomizer.py
snuffysasa: I think I understand now, but this check is gone now and it creates the definition by directly passing member_type to type, So I think we are good?

default_weights.toml
```diff
@@ -31,6 +31,7 @@ perm_duplicate_assignment = 5
 perm_chain_assignment = 5
 perm_long_chain_assignment = 3
 perm_pad_var_decl = 1
+perm_inline_get_structmember = 1
```
simonlindholm, line 34: this can be 0 for ido btw, which doesn't support the `inline` keyword

src/randomizer.py
```diff
@@ -493,6 +493,29 @@ def rec(block: Block) -> None:
...
 
 
+def get_noncolliding_fn_name(ast: ca.FileAST, fn_name: str) -> str:
+
+    used_names: Set[str] = set()
+
+    for item in ast.ext:
+        if (
+            isinstance(item, ca.Decl)
+            and isinstance(item.type, ca.FuncDecl)
```
simonlindholm, line 503:
```suggestion
```
to generalize this function to not just cover functions

src/randomizer.py
```diff
@@ -493,6 +493,29 @@ def rec(block: Block) -> None:
...
+        ):
+            used_names.add(item.name)
+        if isinstance(item, ca.FuncDef) and item.decl.name:
+            used_names.add(item.decl.name)
+
+    new_fn_name = fn_name
+    counter = 1
+    while new_fn_name in used_names:
+        counter += 1
+        new_fn_name = f"{new_fn_name}{counter}"
```
simonlindholm, line 514:
```suggestion
        new_fn_name = f"{fn_name}{counter}"
```

src/randomizer.py
```diff
@@ -2080,6 +2103,85 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+            ):
+                cands.append(node)
+            self.generic_visit(node)
+
+    Visitor().visit(fn.body)
+    ensure(cands)
+
+    cand = random.choice(cands)
+
+    parentOp: Optional[ca.UnaryOp] = ast_util.find_parent_addressOP(fn.body, cand)
```
simonlindholm, line 2130: no camel casing please

src/randomizer.py
```diff
@@ -2080,6 +2103,85 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+
+    fn_decl = ca.Decl(
+        name=new_inline_fn_name,
+        quals=[],
+        align=[],
+        storage=[],
+        funcspec=["inline"],
+        type=ca.FuncDecl(
+            args=ca.ParamList(
+                [ca.Typename(name=None, quals=[], align=[], type=arg_type)]
```
simonlindholm, line 2157: You want a Decl for the param; a Typename is only for the case where the param does not have a name

src/randomizer.py
```diff
@@ -2080,6 +2103,85 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+    replace_node(
+        fn.body,
+        cand if not parentOp else parentOp,
+        ca.FuncCall(ca.ID(new_inline_fn_name), ca.ExprList([cand.name])),
+    )
+
+    # Now create the new inline function definition
+    arg_type: SimpleType = decayed_expr_type(cand.name, typemap)
+    assert isinstance(arg_type.type, ca.TypeDecl)
+    assert arg_type.type.declname
```
simonlindholm, line 2147: I don't think this is true, and you shouldn't need it

default_weights.toml
snuffysasa: check

src/randomizer.py
snuffysasa: done

src/randomizer.py
snuffysasa: okay sure, I changed it... but I don't understand, typename worked and does have the name

snuffysasa: Decl, just seems like a lot for a param

```
    theParam = ca.Decl(
        name=None,
        quals=[],
        align=[],
        storage=[],
        funcspec=[],
        type=arg_type,
        init=None,
        bitsize=None,
    )
```

simonlindholm: There may be further randomization passes, and if we generate invalid AST's (that just happen to stringify to the right thing) they may break. To combat the verbosity, you could create a function `make_decl(name, type)` (which also calls `set_decl_name(decl)`, and which `insert_decl` can make use of).

src/randomizer.py
```diff
@@ -2080,6 +2099,99 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+
+    theParam = ca.Decl(
```
simonlindholm, line 2150: theParam -> param, and I suggest creating a `make_decl` helper (split out parts of `insert_decl`). `name` ought to be made non-None here. `fn_decl` can also make use of that helper (with either an optional parameter for `funcspec` or setting that member afterwards).

src/randomizer.py
```diff
@@ -2080,6 +2099,99 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+            cands.append(expr)
+
+    ensure(cands)
+
+    cand = random.choice(cands)
+
+    ensure(cands)
+    cand = random.choice(cands)
+
+    parentOp: Optional[ca.UnaryOp] = None
```
simonlindholm, line 2127: no camel casing please

src/randomizer.py
simonlindholm: So, this is for magically picking a parameter name out of thin air. When `cand.name` is a ca.ID, which it previously always was, `expr_type` would find its declaration and return that Decl's type, which would incidentally have a name set on it (because of the weird way in which the AST works for declarations) which matches the ca.ID's name. This isn't something that `expr_type` guarantees will work though, it's very subtle, and it will break for non-ca.ID's... Instead I would suggest creating a custom parameter name (say, `"x"`, or falling back to cand.name.name if cand.name is an ca.ID if you want).

src/randomizer.py
```diff
@@ -2080,6 +2099,99 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+
+    member_type: SimpleType = decayed_expr_type(cand, typemap)
+    member_name = cand.field.name
+
+    new_inline_fn_name = get_noncolliding_name(ast, f"get_{member_name}")
+
+    # Replace the StructRef with a FuncCall to the new inline yet to be created
+    replace_node(
+        fn.body,
+        cand if not parentOp else parentOp,
```
simonlindholm, line 2141:
```suggestion
        parentOp or cand,
```

src/randomizer.py
```diff
@@ -2080,6 +2099,99 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+        ),
+        init=None,
+        bitsize=None,
+    )
+
+    set_decl_name(fn_decl)
+
+    return_expr: Expression = ca.StructRef(
+        name=ca.ID(arg_type.type.declname),
+        type="->",
```
simonlindholm, line 2179:
```suggestion
        type=cand.type,
```
or forbid `cand.type == "."` early on

src/randomizer.py
```diff
@@ -2080,6 +2099,99 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+            isinstance(expr, ca.UnaryOp)
+            and expr.op is "&"
+            and isinstance(expr.expr, ca.StructRef)
+        ):
+            cands.append(expr)
+
+    ensure(cands)
+
+    cand = random.choice(cands)
+
```
simonlindholm, line 2123:
```suggestion
```

snuffysasa: oh whoops! I was editing way too late last night, =)

src/randomizer.py
snuffysasa: thought I had something, that prevented "."   but I guess not and it's added explicitly now

src/ast_util.py
```diff
@@ -234,21 +234,35 @@ def get_block_stmts(block: Block, force: bool) -> List[Statement]:
     return ret
 
 
-def insert_decl(
-    fn: ca.FuncDef, var: str, type: SimpleType, random: Optional[Random] = None
-) -> None:
+def make_decl(
+    name: str,
+    type: Union[SimpleType, ca.FuncDecl],
```
simonlindholm, line 239:
```suggestion
    type: "ca.Type",
    *,
```

src/ast_util.py
```diff
@@ -234,21 +234,35 @@ def get_block_stmts(block: Block, force: bool) -> List[Statement]:
...
+    quals: List[str] = [],
+    align: List[ca.Alignas] = [],
+    storage: List[str] = [],
+    funcspec: List[str] = [],
```
simonlindholm, line 243: It would be preferable to have these be `Optional[...]` and default to None, with `or []` at the place they are used. Reason being that Python evaluates parameter defaults just once instead of once per call, and because lists are mutable that can cause mysterious bugs. Likely wouldn't actually ever come up in practice but it's a good rule to learn.

src/randomizer.py
```diff
@@ -2080,6 +2099,78 @@ def visit_Decl(self, decl: ca.Decl) -> None:
...
+        parent_op or cand,
+        ca.FuncCall(ca.ID(new_inline_fn_name), ca.ExprList([cand.name])),
+    )
+
+    # Now create the new inline function definition
+    arg_type: SimpleType = decayed_expr_type(cand.name, typemap)
+
+    param = ast_util.make_decl("x", arg_type)
+
+    fn_decl = ast_util.make_decl(
```
simonlindholm, line 2147: now that make_decl makes these so short we can imo inline them into the usage

simonlindholm (approved):
Looks good! Just some minor style things

src/ast_util.py
snuffysasa: okay

src/randomizer.py
snuffysasa: I inlined one of them into the usage

PR #143, opened 2022-07-02
cache randomized source using a hash, to save time by skipping compiling and scoring identical sources

snuffysasa:
cache randomized source using a hash, to save time by skipping compiling and scoring identical sources

### DIFF
--- a/src/net/evaluator.py
+++ b/src/net/evaluator.py
@@ -217,25 +217,22 @@ def multiprocess_worker(
 
         permuter = permuters[work.perm_id]
         result = permuter.try_eval_candidate(work.seed)
-        if result:
-            if isinstance(result, CandidateResult) and permuter.should_output(result):
-                permuter.record_result(result)
+        if isinstance(result, CandidateResult) and permuter.should_output(result):
+            permuter.record_result(result)
 
-            # Compress the source within the worker. (Why waste a free
-            # multi-threading opportunity?)
-            if isinstance(result, CandidateResult):
-                compressed_source: Optional[bytes] = None
-                if result.source is not None:
-                    compressed_source = zlib.compress(result.source.encode("utf-8"))
-                setattr(result, "compressed_source", compressed_source)
-                result.source = None
+        # Compress the source within the worker. (Why waste a free
+        # multi-threading opportunity?)
+        if isinstance(result, CandidateResult):
+            compressed_source: Optional[bytes] = None
+            if result.source is not None:
+                compressed_source = zlib.compress(result.source.encode("utf-8"))
+            setattr(result, "compressed_source", compressed_source)
+            result.source = None
 
-            time_us = int((time.time() - time_before) * 10 ** 6)
-            task_queue.put(
-                WorkDone(
-                    perm_id=work.perm_id, id=work.id, time_us=time_us, result=result
-                )
-            )
+        time_us = int((time.time() - time_before) * 10 ** 6)
+        task_queue.put(
+            WorkDone(perm_id=work.perm_id, id=work.id, time_us=time_us, result=result)
+        )
 
 
 def read_loop(task_queue: "Queue[Task]", port: Port) -> None:
--- a/src/net/evaluator.py
+++ b/src/net/evaluator.py
@@ -217,22 +217,25 @@ def multiprocess_worker(
 
         permuter = permuters[work.perm_id]
         result = permuter.try_eval_candidate(work.seed)
-        if isinstance(result, CandidateResult) and permuter.should_output(result):
-            permuter.record_result(result)
+        if result:
+            if isinstance(result, CandidateResult) and permuter.should_output(result):
+                permuter.record_result(result)
 
-        # Compress the source within the worker. (Why waste a free
-        # multi-threading opportunity?)
-        if isinstance(result, CandidateResult):
-            compressed_source: Optional[bytes] = None
-            if result.source is not None:
-                compressed_source = zlib.compress(result.source.encode("utf-8"))
-            setattr(result, "compressed_source", compressed_source)
-            result.source = None
+            # Compress the source within the worker. (Why waste a free
+            # multi-threading opportunity?)
+            if isinstance(result, CandidateResult):
+                compressed_source: Optional[bytes] = None
+                if result.source is not None:
+                    compressed_source = zlib.compress(result.source.encode("utf-8"))
+                setattr(result, "compressed_source", compressed_source)
+                result.source = None
 
-        time_us = int((time.time() - time_before) * 10 ** 6)
-        task_queue.put(
-            WorkDone(perm_id=work.perm_id, id=work.id, time_us=time_us, result=result)
-        )
+            time_us = int((time.time() - time_before) * 10 ** 6)
+            task_queue.put(
+                WorkDone(
+                    perm_id=work.perm_id, id=work.id, time_us=time_us, result=result
+                )
+            )
 
 
 def read_loop(task_queue: "Queue[Task]", port: Port) -> None:
--- a/src/permuter.py
+++ b/src/permuter.py
@@ -163,7 +163,6 @@ class Permuter:
         return self._need_all_sources or self.should_output(result)
 
     def _eval_candidate(self, seed: int) -> CandidateResult:
-        t0 = time.time()
 
         # Determine if we should keep the last candidate.
         # Don't keep 0-score candidates; we'll only create new, worse, zeroes.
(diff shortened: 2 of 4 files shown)


src/net/evaluator.py
```diff
@@ -217,22 +217,25 @@ def multiprocess_worker(
...
-            if result.source is not None:
-                compressed_source = zlib.compress(result.source.encode("utf-8"))
-            setattr(result, "compressed_source", compressed_source)
-            result.source = None
-
-        time_us = int((time.time() - time_before) * 10 ** 6)
-        task_queue.put(
-            WorkDone(perm_id=work.perm_id, id=work.id, time_us=time_us, result=result)
-        )
+        if result:
```
simonlindholm, line 220: we do actually need to send something across the network in this case

src/permuter.py
```diff
@@ -261,10 +273,12 @@ def seed_iterator(self) -> Iterator[int]:
             return itertools.repeat(self._force_seed)
         return iter([self._force_seed])
 
-    def try_eval_candidate(self, seed: int) -> EvalResult:
+    def try_eval_candidate(self, seed: int) -> Optional[EvalResult]:
```
simonlindholm, line 276: EvalResult is already an "either a result or an error", wrapping it in an additional layer of "might not exist" feels complex/wrong. Can we either extend EvalResult with an additional enum kind, or save the result in the cache and return that?

src/net/evaluator.py
snuffysasa: thought that might be the case

src/permuter.py
snuffysasa: you're right those options make more sense.
I can do one of those options if you want.  
But how do you feel about me changing it back to EvalResult (Not Optional)
and doing this:
`return EvalError(exc_str="Duplicate Source from Randomization - not scoring", seed=self._cur_seed)`

simonlindholm (changes requested):
I think it's wrong to use EvalError for this. Note that compile errors don't use EvalError either in the normal case; _CompileFailure is only for `--show-errors` mode (I think because we don't want to remove the .o, and thus we don't want to score it which has that as a side effect? So we throw a kind of "internal error" instead of handle that specially further on. It's kinda overcomplicated, ought to be refactored/documented).

A useful consideration: when we encounter a duplicate source, we need to know whether we should have a chance to continue randomizing it or reset because there's a compile error. Just a "duplicate source" signal with no additional information doesn't tell us that.

I can think of two possible alternatives for how to approach this:
- cache a source -> CandidateResult mapping
- wrap randomization in a while loop; keep randomizing while the source is seen and don't expose duplicates outwards
I think I prefer the second option.

Another factor we should consider is memory usage: the way you have it right now has memory growing forever, which could maybe cause a computer to lock up if you run the permuter running overnight on a lot of cores (I haven't done the math). I think the right solution to that is to stop the set growing once it hits some size limit (in theory we'd want an LRU cache, but since permutation is currently stateless there's no need). Could maybe use a bloom filter, too!

snuffysasa:
> I think it's wrong to use EvalError for this. Note that compile errors don't use EvalError either in the normal case; _CompileFailure is only for `--show-errors` mode (I think because we don't want to remove the .o, and thus we don't want to score it which has that as a side effect? So we throw a kind of "internal error" instead of handle that specially further on. It's kinda overcomplicated, ought to be refactored/documented).
> 
> A useful consideration: when we encounter a duplicate source, we need to know whether we should have a chance to continue randomizing it or reset because there's a compile error. Just a "duplicate source" signal with no additional information doesn't tell us that.
> 
> I can think of two possible alternatives for how to approach this:
> 
>     * cache a source -> CandidateResult mapping
> 
>     * wrap randomization in a while loop; keep randomizing while the source is seen and don't expose duplicates outwards
> 
> 
> I think I prefer the second option.
> 
> Another factor we should consider is memory usage: the way you have it right now has memory growing forever, which could maybe cause a computer to lock up if you run the permuter running overnight on a lot of cores (I haven't done the math). I think the right solution to that is to stop the set growing once it hits some size limit (in theory we'd want an LRU cache, but since permutation is currently stateless there's no need). Could maybe use a bloom filter, too!

I did the latter.  It seems to work well.  But it breaks profile time for stringify a candidate... I don't know how much you care to keep that metric for profiling

snuffysasa (commented):
I completed the requested changes

simonlindholm (commented):
Can you fix stringification time? Requires moving away from t0..t3 and computing the times only at the end but that shouldn't be a problem

simonlindholm:
> Another factor we should consider is memory usage: the way you have it right now has memory growing forever, which could maybe cause a computer to lock up if you run the permuter running overnight on a lot of cores (I haven't done the math). I think the right solution to that is to stop the set growing once it hits some size limit (in theory we'd want an LRU cache, but since permutation is currently stateless there's no need). Could maybe use a bloom filter, too!

also please do this

snuffysasa:
> Can you fix stringification time? Requires moving away from t0..t3 and computing the times only at the end but that shouldn't be a problem

I am not sure sure the approach you want me to take.

The stringification will now happen multiple times ` ast_util.to_c(self.ast)`
Do you want me to sum all the times the stringification happens? and find the sum? or an average? or just the first one?

and what about the hashing, and searching the set, should that be counted anywhere? in the randomization? in the stringification?  or make a new one?

src/permuter.py
```diff
@@ -192,11 +195,18 @@ def _eval_candidate(self, seed: int) -> CandidateResult:
 
         # Randomize the candidate
         if self._permutations.is_random():
-            self._cur_cand.randomize_ast()
+            found_new_source = False
+            while not found_new_source:
```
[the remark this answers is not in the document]
snuffysasa: okay, I changed it

snuffysasa:
> > Another factor we should consider is memory usage: the way you have it right now has memory growing forever, which could maybe cause a computer to lock up if you run the permuter running overnight on a lot of cores (I haven't done the math). I think the right solution to that is to stop the set growing once it hits some size limit (in theory we'd want an LRU cache, but since permutation is currently stateless there's no need). Could maybe use a bloom filter, too!
> 
> also please do this

Okay I will look into memory usage and limiting it.

simonlindholm:
>  Do you want me to sum all the times the stringification happens? and find the sum? or an average? or just the first one?

The sum.

> and what about the hashing, and searching the set, should that be counted anywhere? in the randomization? in the stringification? or make a new one?

It's probably negligible time-wise, I'd say throw it in with the stringification.

snuffysasa:
> > Do you want me to sum all the times the stringification happens? and find the sum? or an average? or just the first one?
> 
> The sum.
> 
> > and what about the hashing, and searching the set, should that be counted anywhere? in the randomization? in the stringification? or make a new one?
> 
> It's probably negligible time-wise, I'd say throw it in with the stringification.

I updates the profiling timers.  Although the behavior will be different since in some cases it will just do one randomization pass and stringify, and sometimes it will do several.   So the time for perm and stringify will go up, but compile time and scoring will go down relatively because they are having less iterations

simonlindholm (commented):
For the timing I'd prefer (not that it matters hugely in practice) if you kept a single running timer started at the top of the function and did basically `t_start = t_end` everywhere instead of `t_start = time.time()`. I think a helper timer class could make it looks less verbose (currently the randomization loop is very noisy), something like:
```py
class Timer:
    def __init__(self) -> None:
        self._time = time.time()

    def elapsed(self) -> float:
        """Return time elapsed since construction or last call to `elapsed`."""
        cur = self._time
        self._time = time.time()
        return self._time - cur
```
probably with time.monotonic() instead of time.time() now that I look at docs.

snuffysasa:
Okay so added that if check for length of the set to prevent it from growing too big.

and I added the timer class, but to be honest I wasn't sure I understood your instructions on that, and I am not sure if its cleaner this way... but that's how it is now

simonlindholm:
you misunderstood slightly: my idea was to use a single Timer object for the whole function instead of creating new ones all the time

snuffysasa:
I am having trouble seeing how this would work. The ways I am imaging what you are saying to do results in much more code, which I don't think is what you intend.

snuffysasa:
I am going to revert my previous commit and then add back just the if check on seen sources size

snuffysasa:
I did that.  I request that add the Timer class on your own because I feel this might be a case where its easier for you to do then to explain exactly what you intend.   Or if you do give more explanation I will try to update the PR again.  Or just take it as is.

simonlindholm:
Done in 3be555f1ed3351f1da2c2ab59dd0eb5dd730bb30.

snuffysasa:
> Done in [3be555f](https://github.com/simonlindholm/decomp-permuter/commit/3be555f1ed3351f1da2c2ab59dd0eb5dd730bb30).

I see!  Great

PR #144, opened 2022-07-04
default_to_not_stripping_fn_defs, flag to re-enable

snuffysasa:
default_to_not_stripping_fn_defs, flag to re-enable


src/ast_util.py
```diff
@@ -100,13 +100,15 @@ def visit_If(self, n: ca.If) -> str:
         return super().visit_If(n2)  # type: ignore
 
 
-def extract_fn(ast: ca.FileAST, fn_name: str) -> Tuple[ca.FuncDef, int]:
+def extract_fn(
+    ast: ca.FileAST, fn_name: str, strip_other_fn_defs: bool
```
ethteck, line 104:
```suggestion
    ast: ca.FileAST, fn_name: str, strip_other_fn_defs: bool = True
```
I think it'd make sense to give this arg a default value so it's clear what exactly the default behavior is. I'm a bit confused about this PR becuase I see you passing True in sometimes and False others, so it seems like a behavioral change as opposed to merely the introduction of a new option.

For IDO, I know sometimes function defs are required and affect codegen, but for GCC this isn't the case. I'm thinking it's okay to change default behavior based on the compiler provided in the settings file, but we probably should at least use that information instead of just changing the default behavior for everyone.

ethteck (changes requested):

simonlindholm:
IIRC the context around this might have been to be able to permute inline functions? There's some discussion in Discord if you search for it. I think it stalled waiting for me to make some kind of decision about what a reasonable default behavior would be, but I never came around to that... When using import.py and having that strip away all other functions this PR doesn't have any effect, but it does intentionally change the behavior of non-import.py flows.

PR #145, opened 2022-07-20
alternate solution for supporting windows or systems without pynacl

snuffysasa:
alternate solution for supporting windows or systems without pynacl

Let me know what changes you would like or additions, 
then if this works I will close the other PR

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -25,8 +25,10 @@ from .helpers import (
     static_assert_unreachable,
     trim_source,
 )
-from .net.client import start_client
-from .net.core import ServerError, connect, enable_debug_mode, MAX_PRIO, MIN_PRIO
+
+MIN_PRIO = 0.01
+MAX_PRIO = 2.0
+
 from .permuter import (
     EvalError,
     EvalResult,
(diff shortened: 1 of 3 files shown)


src/main.py
```diff
@@ -401,6 +403,10 @@ def run_inner(options: Options, heartbeat: Callable[[], None]) -> List[int]:
         # Connect to network and create client threads and queues.
         net_conns: "List[Tuple[threading.Thread, Queue[Task]]]" = []
         if options.use_network:
+
+            from .net.client import start_client
+            from .net.core import ServerError, connect, enable_debug_mode
```
simonlindholm, line 408: These imports being here needs a comment, and the README needs to be updated to mention that the `pynacl` dependency is optional and doesn't work on Windows. And I do think we should have a try-except here with a nice error message, otherwise this optional support looks like a bug.

simonlindholm: on second thought I guess it's fine to skip the try-except, given instructions in the readme

snuffysasa:
Okay I made these tweaks

simonlindholm (commented):
Sorry for being super slow to get to this

PR #146, opened 2022-07-24
Randomization pass: Flatten fields out of a substruct

abaresk:
Flattens fields out of a substruct mostly following [these flattening rules](https://gist.github.com/abaresk/132fe737825debb4d2c20ea92d2e4a48). Note, for simplicity we're not restricting the range for downward flattens.

Example: [original source](https://gist.github.com/abaresk/1188213fe0d72ddce142ff20f2c239b8), [permutations](https://gist.github.com/abaresk/620f69b4786de817902bdf9ce7b7caf6)

### DIFF
--- a/src/ast_types.py
+++ b/src/ast_types.py
@@ -199,7 +199,6 @@ def decayed_expr_type(expr: ca.Node, typemap: TypeMap) -> SimpleType:
 def resolve_struct_def(struct: ca.TypeDecl, typemap: TypeMap) -> Optional[ca.Struct]:
     """Resolve struct definition for a given type.
     """
-    # print('resolve structdef of:', struct)
     if isinstance(struct.type, ca.IdentifierType):
         idType = struct.type.names[0]
         if idType in typemap.typedefs:
--- a/src/ast_types.py
+++ b/src/ast_types.py
@@ -8,7 +8,7 @@ They make a number of simplifying assumptions:
 For the purposes of the randomizer these restrictions are acceptable."""
 
 from dataclasses import dataclass, field
-from typing import Union, Dict, Set, List
+from typing import Union, Dict, Set, List, Optional
 
 from pycparser import c_ast as ca
 
@@ -195,6 +195,21 @@ def expr_type(node: ca.Node, typemap: TypeMap) -> Type:
 def decayed_expr_type(expr: ca.Node, typemap: TypeMap) -> SimpleType:
     return pointer_decay(expr_type(expr, typemap), typemap)
 
+
+def resolve_struct_def(struct: ca.TypeDecl, typemap: TypeMap) -> Optional[ca.Struct]:
+    """Resolve struct definition for a given type.
+    """
+    # print('resolve structdef of:', struct)
+    if isinstance(struct.type, ca.IdentifierType):
+        idType = struct.type.names[0]
+        if idType in typemap.typedefs:
+            return resolve_struct_def(typemap.typedefs[idType], typemap)
+        return None
+    if struct.type.decls is not None:
+        return struct.type
+    if struct.type.name in typemap.struct_defs:
+        return typemap.struct_defs[struct.type.name]
+    return None
 
 def same_type(
     type1: Type, type2: Type, typemap: TypeMap, allow_similar: bool = False
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -202,10 +202,9 @@ def struct_field_idx(struct: ca.Struct, field_name: str) -> int:
 
 
 def struct_substruct_idx(parent: ca.Struct, substruct: ca.Struct, typemap: TypeMap) -> int:
-    print('parent:')
     parent.show()
     for i, decl in enumerate(parent.decls):
-        if isinstance(decl.type, ca.TypeDecl): # and same_type(decl.type.type, substruct, typemap):
+        if isinstance(decl.type, ca.TypeDecl):
             decl_struct = resolve_struct_def(decl.type, typemap)
             if decl_struct is substruct:
                 return i
@@ -349,18 +348,10 @@ def flatten_fields(parent: ca.Struct, parent_field: str, substruct: ca.Struct, s
     Returns a mapping from the original names of the moved fields to their
     (potentially) new names in the parent.
     """
-    # print('parent:')
-    # parent.show()
-    # print('substruct:')
-    # substruct.show()
-
     parent_field_names = set([decl.name for decl in parent.decls])
 
     parent_idx = struct_field_idx(parent, parent_field)
     shift_idx = struct_field_idx(substruct, substruct_field)
-    print('parent_idx:', parent_idx)
-    print('shift_idx:', shift_idx)
-    print('flatten upwards', flatten_upwards)
     if flatten_upwards:
         moved_fields = move_fields(substruct, parent, 0, shift_idx + 1, parent_idx)
         substruct_i = parent_idx + shift_idx + 1
@@ -370,14 +361,8 @@ def flatten_fields(parent: ca.Struct, parent_field: str, substruct: ca.Struct, s
 
     # If substruct is empty, remove it from the parent
     if not substruct.decls:
-        print('gloom')
         struct_remove_field_i(parent, substruct_i)
         parent_field_names.remove(parent_field)
-
-    print('newParent:')
-    parent.show()
-    print('newSubstruct:')
-    substruct.show()
 
     return deduplicate_struct_field_names(moved_fields, parent_field_names)
 
--- a/src/ast_util.py
+++ b/src/ast_util.py
@@ -217,59 +217,29 @@ def copy_struct(ast: ca.FileAST, typemap: TypeMap, struct: ca.Struct) -> ca.Stru
     """Adds a new struct with a definition matching `struct` to the AST.
     Returns this new struct.
     """
-    print('copying struct!')
     new_struct = copy.deepcopy(struct)
     base_name = struct.name if struct.name else struct.decls[0].name
     new_struct.name = get_noncolliding_struct_name(typemap, base_name)
-    # TODO NEXT: Add new struct to AST (wrap in a Decl)! Place it in the
-    # top-level, right before typedef/decl containing the base struct. That way,
-    # it has access to all its dependencies.
-    # 
-    # NOTE TO SELF: It can't be the final struct definition, because the parent
-    # is defined before. It can't be first, in case it depends on other structs.
-    # 
-    # CODE REVIEW NOTE: Make sure that this new definition gets deleted after the
-    # randomization pass! Honestly, I'm not sure how to do this cleanly...
-    # 
-    # It's not a big deal now. It's rare to have to add a new struct for this
-    # pass. It will be necessary for the NEST rule, otherwise you'll have a
-    # bunch of new struct defs polluting candidates down the line.
-    #
-    # Maybe, if you have to add a new struct, you should just deeply copy the
-    # entire AST for that candidate?
-    #
-    # TODO: Add the new struct definition in such a way that it doesn't persist
-    # across candidates.
-    #   Nvm, this doesn't seem to persist across candidates...
-    print('new struct name:', new_struct.name)
+
+    # Insert the new struct right above the base struct. This preserves any
+    # struct field dependencies.
     insert_idx = ast_ext_struct_idx(ast, struct)
-    print(f'inserting in ast.ext[{insert_idx}]:')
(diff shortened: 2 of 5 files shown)


src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+    used_names: Set[str] = set()
+    for item in ast.ext:
+        if isinstance(item, ca.Decl) and item.name:
+            used_names.add(item.name)
+        if isinstance(item, ca.FuncDef) and item.decl.name:
+            used_names.add(item.decl.name)
+    return get_noncolliding_name(used_names, name)
+
+
+def get_noncolliding_struct_name(typemap: TypeMap, name: str) -> str:
```
simonlindholm, line 179: Instead of two different functions I'd suggest a single one that works for both struct and function names (and variables, ...). Function names and typedefs share the same namespace anyway.

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+    used_names: Set[str] = set()
+    for struct_name in typemap.struct_defs.keys():
+        used_names.add(struct_name)
+    for type_name in typemap.typedefs.keys():
+        used_names.add(type_name)
```
simonlindholm, line 184:
```suggestion
    used_names: Set[str] = {
        *typemap.struct_defs.keys(),
        *typemap.typedefs.keys(),
    }
```

src/ast_types.py
```diff
@@ -196,6 +196,20 @@ def decayed_expr_type(expr: ca.Node, typemap: TypeMap) -> SimpleType:
     return pointer_decay(expr_type(expr, typemap), typemap)
 
 
+def resolve_struct_def(struct: ca.TypeDecl, typemap: TypeMap) -> Optional[ca.Struct]:
+    """Resolve struct definition for a given type.
+    """
+    if isinstance(struct.type, ca.IdentifierType):
+        idType = struct.type.names[0]
+        if idType in typemap.typedefs:
+            return resolve_struct_def(typemap.typedefs[idType], typemap)
```
simonlindholm, line 205: This looks like it shouldn't type-check: `typemap.typedefs[...]` is a `Type`, while `resolve_struct_def` accepts a ca.TypeDecl. Did you run `mypy`?

Slight shortening (though will need some isinstance TypeDecl check to work with mypy):
```suggestion
    struct = resolve_typedefs(struct, typemap)
    if isinstance(struct.type, ca.IdentifierType):
```

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+    class StructRefVisitor(ca.NodeVisitor):
+        def visit_StructRef(self, node: ca.StructRef) -> None:
+            nonlocal chain
+            if not isinstance(node.name, ca.StructRef) or node.type != ".":
+                chain = 0
+                return
+            self.generic_visit(node)
+            chain += 1
+
+    StructRefVisitor().visit(node)
```
simonlindholm, line 285:
```suggestion
    while isinstance(node.name, ca.StructRef) and node.type == ".":
        chain += 1
        node = node.name
```
Using a visitor confused me for a second, makes it look like you're visiting an entire AST.

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+
+def count_chained_structrefs(node: ca.StructRef) -> int:
+    """Count the number of chained StructRef's that are required to reach the
+    field in `node`.
+
+    Examples:
+      - foo            --> 0
+      - bar.foo        --> 1
+      - baz.bar.foo    --> 2
+      - bar->foo       --> 0
```
simonlindholm, line 272: I'm getting 1 for this trying to trace the code: the code looks at `node.type` instead of `node.name.type` so it does traverse all the way to the -> node before bailing, adding +1 along the way. Guessing that's unintentional?

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+      - baz.bar[i].foo --> 0
```
simonlindholm, line 273: and I'm getting 1 for this, which *is* the number of trailing `.blah`'s so I think that 1 is actually correct?

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+def struct_remove_field(struct: ca.Struct, field_idx: int) -> None:
+    struct.decls = struct.decls[:field_idx] + struct.decls[field_idx+1:]
+
+
+def count_chained_structrefs(node: ca.StructRef) -> int:
+    """Count the number of chained StructRef's that are required to reach the
+    field in `node`.
+
+    Examples:
+      - foo            --> 0
```
simonlindholm, line 269: `foo` isn't a ca.StructRef, so this doesn't match the function signature. If you mean `foo.something` I agree that that yields 0 with the current code. (But having the function accept `node: Expression` maybe makes more sense?)

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+    """
+    for i, decl in enumerate(parent.decls):
+        if isinstance(decl.type, ca.TypeDecl):
+            decl_struct = resolve_struct_def(decl.type, typemap)
+            if decl_struct is substruct:
+                return i
+    return -1
+
+
+def struct_rename_field(field: ca.Decl, name: str) -> None:
```
simonlindholm, line 251: This is the same as `set_decl_name` more or less

src/randomizer.py
```diff
@@ -2298,6 +2286,61 @@ def perm_inline(
...
+# A map of all the parents this substruct is referenced from to the list of
+# references via that parent.
+ParentRefs = Dict[ca.Struct, List[ca.StructRef]]
+
+def perm_flatten_struct(
+    fn: ca.FuncDef, ast: ca.FileAST, indices: Indices, region: Region, random: Random
+) -> None:
+    """Moves fields from a nested struct into its parent struct. Only types used
+    within the given region are affected."""
+    ast_util.deepcopy_struct_defs(ast)
```
simonlindholm, line 2298: The AST must not be modified until we're sure that no RandomizationFailure exception is thrown, or bad things will happen

simonlindholm (commented):
Some preliminary comments, haven't got to the actually interesting parts yet

src/randomizer.py
```diff
@@ -2298,6 +2286,61 @@ def perm_inline(
...
+    shift_idx = random.randint(0, len(substruct.decls)-1)
+    flatten_upwards = random_bool(random, PROB_FLATTEN_UPWARDS)
+
+    # Move fields out of substruct's struct definition and into the parent's
+    # struct definition.
+    moved_names = ast_util.flatten_fields(parent, substruct, shift_idx, flatten_upwards, typemap)
+
+    # Commit the changes to all StructRefs under this parent. We should only
+    # update references to fields that were moved.
+    for structRef in substructs[old_substruct][parent]:
```
simonlindholm, line 2338: nit: struct_ref

src/ast_util.py
```diff
@@ -166,6 +166,126 @@ def generic_visit(self, node: ca.Node) -> None:
...
+    """Returns the index of the node in `ast.ext` that contains the definition
+    of `struct`.
+    """
+    found = False
+    class StructVisitor(ca.NodeVisitor):
+        def visit_Struct(self, node: ca.Struct) -> None:
+            nonlocal found
+            if node is struct:
+                found = True
+            self.generic_visit(node)
```
simonlindholm, line 207: could make this into a general "contains node" method

src/ast_types.py
abaresk: Done + mypy fixes

src/ast_util.py
abaresk: I split the functions by namespace

src/ast_util.py
abaresk: obsolete

src/ast_util.py
abaresk: Good idea! Added

src/ast_util.py
abaresk: Ah thanks, I switched to `set_decl_name()` here

src/ast_util.py
abaresk: Good point, I switched to the `Expression` argument type

src/ast_util.py
abaresk: I think this actually should be 0, following the new implementation. This expression maps to
```
StructRef: ->
  ID: bar
  ID: foo
```

Here, `expr.type` will be "->", so we'll never enter the loop.

src/ast_util.py
abaresk: Ah yep, with the new implementation, this should be 1. For extra context, this expression maps to
```
StructRef: .
  ArrayRef: 
    StructRef: .
      ID: baz
      ID: bar
    ID: i
  ID: foo
```

src/ast_util.py
abaresk: Done

src/ast_util.py
```diff
@@ -253,6 +373,53 @@ def visit_Assignment(self, _: ca.Node) -> None:
...
+
+def flatten_fields(parent: ca.Struct, substruct: ca.Struct, shift_idx: int, flatten_upwards: bool, typemap: TypeMap) -> Dict[str, str]:
+    """Flatten fields from `substruct` into `parent`. If `flatten_upwards` is
+    true, flatten `field` and all fields above upwards into `parent`. Otherwise,
+    flatten `field` and all fields below downwards into `parent`.
+
+    Returns a mapping from the original names of the moved fields to their
+    (potentially) new names in the parent.
+    """
+    parent_field_names = set([decl.name for decl in parent.decls])
```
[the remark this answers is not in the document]
abaresk: Done

src/randomizer.py
```diff
@@ -2298,6 +2286,61 @@ def perm_inline(
     ast.ext.insert(ast.ext.index(fn), fn_def)
 
 
+# A map of all the parents this substruct is referenced from to the list of
+# references via that parent.
+ParentRefs = Dict[ca.Struct, List[ca.StructRef]]
```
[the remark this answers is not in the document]
abaresk: Done

src/randomizer.py
abaresk: To summarize what we discussed offline, we will run into issues if the function definition AST node is touched. You suggested asserting that the function node isn't replaced.

I considered adding this assertion inside `deepcopy_struct_defs()`. I feel like this might not be necessary though. By definition, `deepcopy_struct_defs()` won't be touching the function node.

This randomization pass doesn't modify anything inside the function node until the every end, after it has ensured there won't be a RandomizationFailure. So I think we should be okay -- please let me know though if I missed something or you feel otherwise.

src/randomizer.py
```diff
@@ -2298,6 +2286,61 @@ def perm_inline(
...
+    within the given region are affected."""
+    ast_util.deepcopy_struct_defs(ast)
+    typemap = build_typemap(ast, fn)
+    substructs: Dict[ca.Struct, ParentRefs] = defaultdict(lambda: defaultdict(list))
+
+    # Get all StructRef's in the region that are within a substruct. The fields
+    # must be accessed from a parent struct.
+    class StructRefVisitor(ca.NodeVisitor):
+        def visit_StructRef(self, node: ca.StructRef) -> None:
+            if region.contains_node(node) and ast_util.count_chained_structrefs(node) >= 1:
```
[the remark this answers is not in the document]
abaresk: Ah yes, thanks for catching this!

src/randomizer.py
```diff
@@ -2298,6 +2286,61 @@ def perm_inline(
...
+    # copy of `substruct` in order to leave the other parents unaffected.
+    old_substruct = substruct
+    if len(substructs[substruct].keys()) > 1:
+        substruct_idx = ast_util.struct_substruct_idx(parent, substruct, typemap)
+        substruct = ast_util.copy_struct(ast, typemap, substruct)
+        ast_util.struct_set_field_type(parent, substruct_idx, substruct)
+        typemap = build_typemap(ast, fn)
+
+    ensure(substruct.decls)
+    shift_idx = random.randint(0, len(substruct.decls)-1)
```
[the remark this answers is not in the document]
abaresk: Done

src/randomizer.py
abaresk: Done

abaresk:
Sorry for addressing the comments so late -- haven't had as much free time lately :/

PR #149, opened 2022-09-05
Don't strip sections from asm, convert late_rodata to rodata, add dlabel to prelude

ethteck:


import.py
```diff
@@ -66,23 +66,14 @@ def parse_asm(asm_file: str) -> Tuple[str, str]:
...
-                    cur_section = line.strip()
-                if cur_section == ".text":
-                    if func_name is None and line.strip().startswith("glabel "):
-                        func_name = line.split()[1]
-                    asm_lines.append(line)
+                # .late_rodata is non-standard asm, so change it to .rodata
+                if line.strip() == ".section .late_rodata":
+                    line = ".section .rodata"
+
+                if func_name is None and line.strip().startswith("glabel "):
```
simonlindholm, line 74: some decomps use glabel for data, so I think you need to maintain the section check for this

PR #151, opened 2022-10-06
Allow .globl as function label prefix

mkst:
any thoughts / objections? is there a better way to do this? psx assembler is not a fan of macros :(


simonlindholm:
Happy to take this. The alternative would be keep two sets of asm files, one set for use with INCLUDE_ASM and one for permuter/m2c. I suspect going the route of a single asm format will require some more changes, including to permuter's import.py to have it not use prelude.inc and shell out to gnu as as the assembler.

PR #157, opened 2023-03-07
remove trailing nops from objdump output

ethteck:
image
hopefully prevents bugs like this from surfacing in the future

not sure if explicitly checking for "nop" is arch agnostic enough, but also not sure if that's worth addressing here

### DIFF
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -416,6 +416,10 @@ def simplify_objdump(
             nops = 0
             output_lines.append(Line(row=row, has_symbol=False, mnemonic=mnemonic))
 
+    # Remove trailing nops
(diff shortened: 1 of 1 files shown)


simonlindholm:
Can you remove the part just above too where it does
```py
        if row == "nop":
            # strip trailing nops; padding is irrelevant to us
            nops += 1
        else:
            for _ in range(nops):
                output_lines.append(Line(row="nop", has_symbol=False, mnemonic="nop"))
            nops = 0
            output_lines.append(Line(row=row, has_symbol=False, mnemonic=mnemonic))
```
and just do `output_lines.append(Line(row=row, has_symbol=False, mnemonic=mnemonic))`?

Issue #154, opened 2023-01-30
Permute improvements


ethteck:
The permuter should be able to work off of improvements it has already found, deepening the search space. Ideally this would work when running multithreaded.

Each `Permuter` would have an immutable starting point, and main.py or something else would create new `Permuter`s when improvements to a base are found. We also need to worry about caching results between Permuters to avoid storing the same improvements multiple times. (simon to fill in details)

One pitfall of this approach is that it's sure to generate tons of noisy mutations to the original code, especially with each new layer of findings. To address that, we should try to eliminate meaningless code mutations.

simonlindholm:
> We also need to worry about caching results between Permuters to avoid storing the same improvements multiple times. (simon to fill in details)

Basically, `Permuter.hashes` for all the Permuters needs to point to the same set. I think we may get away with just making the Permuter constructor accept `hashes` as an argument.

There's some subtleties around threading here: what happens is the parent keeps a Permuter per directory you pass in on the cmdline, and each worker keeps a copy of all the Permuters. For each candidate result found by a Permuter, it checks `should_output()` to see if it should be outputted, and if so runs `record_result()` which puts it in the hash set so it won't be found again in the future (unless score = 0 because multiple score = 0 matches is fine). This is done in each worker as a prefilter and again in the parent process if the worker's filter matches. The prefilter is an optimization which avoids needing to serialize AST into C + send it across process (/network) boundaries for each candidate. I don't think this is actually something you need to be aware of; if all `Permuter.hashes` point to the set same the pickling done when starting the multiprocessing workers will ensure that happens in the workers too.

I see a couple of possible ways forward for this:

1. have Permuter grow some field for possible nice starting points, and remember which point was started from for the current iteration to know what score is considered good.
2. create a new wrapper class that will take Permuter's place, and keep a list of old-Permuters as starting points and forward calls to them.
3. use a separate Permuter for each starting point, have main.py manage creation/deletion of Permuters on the fly.
4. give up on a wholly automatic solution and pick out the starting points on startup instead, managing them as separate Permuters with shared `hashes`.

4 feels clearly easiest. UI is a bit unclear, but could perhaps have a flag for "in addition to the starting point, also try to start from the top N entries", or maybe some way of manually picking out which ones look reasonable to start from.

3 is pretty terrible, especially with how messy the scheduler code in main.py is (some of which is permuter@home's fault).

[...]

PR #158, opened 2023-03-23
Spawn new Permuters when improvements are found

ethteck:
This is very much a WIP. I've used it to match some stuff for PM already, but it can't be merged as-is because I don't know how to properly do multithreading in Python. There are 3 main issues I see:

1. I'm not sure if the "new permuter queue for each process" approach I have is ideal, and I'm not sure if the things I lifted into the context class are appropriate there. This is more of a design concern.
2. Quite often, right after a new Permuter is created, the worker proceses get an incoming task for the new permuter before they have added it to their internal list, so they are unable to run the task. I don't know how to write the permuter pass-off mechanism to avoid this scenario. For now I'm just discarding the task.
3. As the process goes and more sub-Permuters are created, there seems to be a considerable slowdown. I suspect this might be related to 2. but I'm not sure. 
4. I haven't really considered p@h yet, and I don't know how this will fit into it.

To-dos:
- [ ] Fix bug (2.) with task that we can't complete
- [ ] Fix gradual slowdown (3.)
- [ ] Add a constructor for Permuter which takes an existing Permuter and new source / a new AST and which doesn't log startup info to console
- [ ] Add args that influence when a new Permuter is created

Open to all and any suggestions for how to write this better.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -98,6 +98,14 @@ class EvalContext:
     errors: int = 0
     overall_profiler: Profiler = field(default_factory=Profiler)
     permuters: List[Permuter] = field(default_factory=list)
+    seed_iterators: List[Optional[Iterator[int]]] = field(default_factory=list)
+    new_permuter_queues: List["Queue[Permuter]"] = field(default_factory=list)
+
+    def add_permuter(self, permuter: Permuter) -> None:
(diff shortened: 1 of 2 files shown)


ethteck:
fixes #154

PR #164, opened 2023-10-01
Allow passing arguments to main() from external source

bismurphy:
Hi there! Thanks for making decomp-permuter, I've found it immensely helpful.

I'd like to leverage `import.py` for use in my own project. Specifically, I'm trying to make a Python script that will automate most of the process of creating a decomp.me page from raw Assembly. This includes things like putting rodata into the .asm file prior to the initial decompilation. In order to create the decomp.me page, I'm currently doing:

```
 import_string = f'tools/decomp-permuter/import.py {c_filename} {asm_filename} --decompme'
    print(f"Calling {import_string}")
    os.system(import_string)
```

But this calling out to the OS seems a little clumsy. It would be nice if I could call import.py from within Python itself, by importing import.py and then calling its `main()` function. However, since `main()` runs `argparse` from within, rather than taking arguments, there's no way (that I'm aware of) to call `main()` externally and give it arguments, without messing with `sys.argv` during runtime.

This PR simply moves the argument parsing into the `if __name__ == __main__` block, and then calls `main()` with the arguments that were parsed. This way, `main()` can also be called with args that come from an alternative source, rather than only what `argparse` finds.

Ultimately, this change should not at all affect using `decomp-permuter` as a standalone utility, but should make it much easier to integrate it into other Python workflows.

### DIFF
--- a/import.py
+++ b/import.py
@@ -769,60 +769,7 @@ def write_to_file(cont: str, filename: str) -> None:
         f.write(cont)
 
 
-def main() -> None:
-    parser = argparse.ArgumentParser(
-        description="""Import a function for use with the permuter.
-        Will create a new directory nonmatchings/<funcname>-<id>/."""
-    )
-    parser.add_argument(
-        "c_file",
-        help="""File containing the function.
-        Assumes that the file can be built with 'make' to create an .o file.""",
-    )
-    parser.add_argument(
-        "asm_file_or_func_name",
-        metavar="{asm_file|func_name}",
-        help="""File containing assembly for the function.
-        Must start with 'glabel <function_name>' and contain no other functions.
-        Alternatively, a function name can be given, which will be looked for in
-        all GLOBAL_ASM blocks in the C file.""",
-    )
-    parser.add_argument(
-        "make_flags",
-        nargs="*",
-        help="Arguments to pass to 'make'. PERMUTER=1 will always be passed.",
-    )
-    parser.add_argument(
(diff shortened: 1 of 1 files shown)


simonlindholm:
Sounds like a worthy goal!

I'd like to keep the argparser code in a function still, otherwise `parser` enters global scope and can be accidentally referenced from other functions. I don't like having `args` be part of public API, since new arguments can be added at any point, which would break your code. How about making `main` accept `sys.argv` as an argument and pass that to `parse_args`?

bismurphy:
I'm slightly confused by your final sentence, I think this is just a wording thing.

> How about making `main` accept `sys.argv` as an argument and pass that to parse_args?

I think what you mean is "Rather than having `main` accept an args dictionary (coming from `parser.parse_args()`, make `main` accept an args list (which may come from `sys.argv`, or anywhere else." 

So `if __name__ == __main__` would call `main(sys.argv)`, and external callers like the one I'm making would do something like `import.main([c_filename, asm_filename, '--decompme')`.  

Am I understanding that correctly?

simonlindholm:
Yeah, exactly.

bismurphy:
Excellent idea. Updated! Now the PR changes are incredibly minimal, which is nice to see.

simonlindholm:
Nice! Can you type the argument as `List[str]` to make mypy happy?

bismurphy:
Done!

PR #165, opened 2024-01-19
decomp.me: Use new scratch claim token

encounter:
A change introduced in https://github.com/decompme/decomp.me/pull/945

It's no longer possible to claim a scratch without a token. This new endpoint will automatically claim the scratch when visited.

### DIFF
--- a/import.py
+++ b/import.py
@@ -904,7 +904,7 @@ def main(arg_list: List[str]) -> None:
                 json_data: Dict[str, str] = json.loads(resp)
                 if "slug" in json_data:
                     slug = json_data["slug"]
-                    token = json_data["claim_token"]
+                    token = json_data.get("claim_token")
                     if token:
                         print(f"https://decomp.me/scratch/{slug}/claim?token={token}")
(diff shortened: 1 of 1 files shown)


ethteck (approved):

import.py
```diff
@@ -904,7 +904,11 @@ def main(arg_list: List[str]) -> None:
                 json_data: Dict[str, str] = json.loads(resp)
                 if "slug" in json_data:
                     slug = json_data["slug"]
-                    print(f"https://decomp.me/scratch/{slug}")
+                    token = json_data["claim_token"]
```
mkst, line 907: Do you want `json_data.get("token")` so it'll work before the decomp.me change is deployed?

simonlindholm:
```suggestion
                    token = json_data.get("claim_token")
```

simonlindholm (approved):

import.py
encounter: Thanks, I always forget about KeyError in Python 😅

PR #166, opened 2024-01-19
Fixed MacOS support to work with mips binaries from homebrew

KiritoDv:


src/objdump.py
```diff
@@ -142,7 +152,7 @@ class ArchSettings:
     re_includes_sp=re.compile(r"\b(sp|s8)\b"),
     sp_ref_insns=["addiu"],
     reloc_str="R_MIPS_",
-    objdump=["mips-linux-gnu-objdump", "-drz", "-m", "mips:4300"],
+    objdump=[f"{get_prefix()}-objdump", "-drz", "-m", "mips:4300"],
```
simonlindholm, line 155: This will be called even if MIPS_SETTINGS is not used, which is a problem if no MIPS toolchain is installed. Maybe there should be an objdump_cands property on ArchSettings which can contain all candidates for prefixes on each arch?

import.py
```diff
@@ -651,6 +651,8 @@ def get_decompme_compiler_name(
 
     for path, compiler_name in compiler_mappings.items():
         assert isinstance(compiler_name, str)
+        # This is an annoying hack because decompme doesnt seem to know about macos
+        compiler_path = compiler_path.replace("macos", "linux")
```
simonlindholm, line 655: Hm... The original intent was that you would add another entry to your tools/permuter_settings.toml to handle this. Maybe we can add support for wildcards to that file?

src/objdump.py
KiritoDv: Oh thats a good idea, let me do it

src/objdump.py
```diff
@@ -434,7 +428,13 @@ def fn(pat: Match[str]) -> str:
 def objdump(
     o_filename: str, arch: ArchSettings, *, stack_differences: bool = False
 ) -> List[Line]:
-    output = subprocess.check_output(arch.objdump + [o_filename])
+    executable = None
+    for cand in arch.executable:
+        if shutil.which(cand):
+            executable = cand
+            break
```
simonlindholm, line 435: Can you handle the case where no executable is found by printing some nice error? (Would also be nice to do this detection only at startup, but I guess there's no good place for it, so eh. Maybe we should add an `@lrucache` to the detection function though)

import.py
```diff
@@ -651,9 +652,8 @@ def get_decompme_compiler_name(
 
     for path, compiler_name in compiler_mappings.items():
         assert isinstance(compiler_name, str)
-        # This is an annoying hack because decompme doesnt seem to know about macos
-        compiler_path = compiler_path.replace("macos", "linux")
-        if path == compiler_path:
+
+        if fnmatch.fnmatch(compiler_path, path):
```
simonlindholm, line 656: Oh, convenient! Can you extend the error messages below with " (wildcards allowed)" or something like that?

KiritoDv: Done!

src/objdump.py
KiritoDv: I would love to do it, but im not that familiar with python, its the first time i heard about lrucache 0:

import.py
```diff
@@ -672,7 +674,7 @@ def get_decompme_compiler_name(
     trail = "permuter_settings.toml, where ... is one of: " + ", ".join(available_ids)
     if compiler_mappings:
         print(
-            "Please add an entry:\n\n"
+            "Please add an entry: (wildcards allowed!)\n\n"
```
simonlindholm, line 677: Add it to the print below as well, for the no-pre-existing section case. (I'd make a review suggestion and accept it myself but github only allows making suggestions up to three lines away from the modified code...) Also, maybe move it to the end instead of here after the colon which looks a bit weird to me grammatically?

src/objdump.py
```diff
@@ -424,7 +428,15 @@ def fn(pat: Match[str]) -> str:
...
+    if executable is None:
+        raise Exception(f"Could not find objdump executable for {arch.name}.")
```
simonlindholm, line 437: Add some info about which executable names are searched for to this message, because I would definitely expect this to show up in practice.

src/objdump.py
simonlindholm: It's super simple to use, it just marks a function as "cache the response value and return it if someone passes the same parameters again". Totally fine if you want to leave it out though.

Issue #67, opened 2020-03-24
Use Levenshtein distance instead of Python's difflib


simonlindholm:
The latter does not obey the triangle inequality -- making a function match worse can improve the score by breaking parts that are accidentally the same between source/target and thus making it line instructions up better.

Using Levenshtein similar to https://github.com/simonlindholm/asm-differ/commit/846a069840a20e099f4b7fb7f1d2b06ea5580b95 would fix this.

Not sure how much of a problem this is in practice.

PR #168, opened 2024-02-03
 #67 Add Levenshtein as a diff algorithm option

Tmcg2:
#67
I have tested the local version of these changes, don't know if I got the permuter@home related stuff correct or not.

Cribs very, very heavily off simonlindholm/asm-differ@846a069

Leaves difflib as the default

Motivated by a unique circumstance in the MK64 decomp where the difflib and Levenshtein algorithms disagree ever so slightly, which made it hard to interpret the permuter's output when comparing it to asm-differ's.

### DIFF
--- a/src/scorer.py
+++ b/src/scorer.py
@@ -1,3 +1,4 @@
+import difflib
 import hashlib
 import re
 from typing import Tuple, List, Optional
@@ -28,6 +29,11 @@ class Scorer:
         self.stack_differences = stack_differences
         self.algorithm = algorithm
         self.debug_mode = debug_mode
+        _, self.target_seq = self._objdump(target_o)
+        self.difflib_differ: difflib.SequenceMatcher[str] = difflib.SequenceMatcher(
+            autojunk=False
+        )
+        self.difflib_differ.set_seq2([line.mnemonic for line in self.target_seq])
 
     def _objdump(self, o_file: str) -> Tuple[str, List[Line]]:
         lines = objdump(o_file, self.arch, stack_differences=self.stack_differences)
@@ -37,7 +43,6 @@ class Scorer:
         if not cand_o:
             return Scorer.PENALTY_INF, ""
 
-        _, target_seq = self._objdump(self.target_o)
         objdump_output, cand_seq = self._objdump(cand_o)
 
         num_stack_penalties = 0
@@ -131,12 +136,6 @@ class Scorer:
         if self.algorithm == "levenshtein":
             import Levenshtein
 
-            # I wish I understood what this function actually did
-            # My loose understanding is thus:
-            #     The function converts a list of mnemonics into a single string,
-            #     but the "characters" of the string are not ascii values?
-            # Copied verbatim from:
-            # https://github.com/simonlindholm/asm-differ/commit/846a069840a20e099f4b7fb7f1d2b06ea5580b95
             remapping = dict()
 
             def remap(seq: List[str]) -> str:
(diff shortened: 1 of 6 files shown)


src/scorer.py
```diff
@@ -35,6 +37,7 @@ def score(self, cand_o: Optional[str]) -> Tuple[int, str]:
         if not cand_o:
             return Scorer.PENALTY_INF, ""
 
+        _, target_seq = self._objdump(self.target_o)
```
simonlindholm, line 40: Move this back into the constructor -- we want to avoid objdumping the target repeatedly

src/scorer.py
```diff
@@ -125,28 +128,60 @@ def diff_insert(line: str) -> None:
...
 
-        self.differ.set_seq1([line.mnemonic for line in cand_seq])
-        result_diff = self.differ.get_opcodes()
+        if self.algorithm == "levenshtein":
+            import Levenshtein
+
+            # I wish I understood what this function actually did
+            # My loose understanding is thus:
+            #     The function converts a list of mnemonics into a single string,
+            #     but the "characters" of the string are not ascii values?
```
simonlindholm, line 137: This is exactly correct. The Levenshtein library unfortunately only allows performing diffs on char sequences while we want to diff string sequences, so we map one to the other. As long as we don't exceed 0x110000 different symbols this hack works out...

I suggest removing/rewording this comment because it doesn't quite fit in the code base :) (It's useful context for the PR though. In the past I've done a self-review of the PR to similar effect.)

src/scorer.py
```diff
@@ -16,16 +15,19 @@ class Scorer:
...
         self.target_o = target_o
         self.arch = get_arch(target_o)
         self.stack_differences = stack_differences
+        self.algorithm = algorithm
         self.debug_mode = debug_mode
-        _, self.target_seq = self._objdump(target_o)
-        self.differ: difflib.SequenceMatcher[str] = difflib.SequenceMatcher(
-            autojunk=False
-        )
-        self.differ.set_seq2([line.mnemonic for line in self.target_seq])
```
simonlindholm, line 28: It would be good if we could keep this differ around in the difflib case -- the docs say
```
set_seq2(b) method of difflib.SequenceMatcher instance
    Set the second sequence to be compared.
...    
    SequenceMatcher computes and caches detailed information about the
    second sequence, so if you want to compare one sequence S against
    many sequences, use .set_seq2(S) once and call .set_seq1(x)
    repeatedly for each of the other sequences.
```
and this feels like a performance win we potentially care about

simonlindholm (commented):
Nice! Some minor comments

src/scorer.py
```diff
@@ -125,8 +133,28 @@ def diff_insert(line: str) -> None:
...
             deletions.append(line)
 
-        self.differ.set_seq1([line.mnemonic for line in cand_seq])
-        result_diff = self.differ.get_opcodes()
+        if self.algorithm == "levenshtein":
+            import Levenshtein
+
+            remapping = dict()
+
+            def remap(seq: List[str]) -> str:
```
Tmcg2, line 141: Having slept on it, I think I get why this remapping is necessary.
Given 2 sequences of instruction mnemonics:
```
a=["lw", "sll", "sw"]
b=["lw", "sw", "sw"]
```
I'd be tempted to do:
```
result = Levensthein.op_codes("".join(a), "".join(b))
```
But that'll cause the perceived diff to be larger than one. We want to know the number mnemonics that are different, not the number of characters that differ between the lists of mnemonics.

So this function gives each mnemonic a unique number (in a pretty novel fashion) and converts that number directly into ascii. Now the Levenshtein algorithm will give us back a diff count that matters for our purposes

PR #170, opened 2024-05-13
Add asm_prelude_file setting for import.py

cadmic:
The default prelude.inc assumes MIPS. I'm not sure if there's already a way to turn that off, but having a project-wide setting seems useful

### DIFF
--- a/import.py
+++ b/import.py
@@ -710,7 +710,7 @@ def write_compile_command(compiler: List[str], cwd: str, out_file: str) -> None:
     os.chmod(out_file, 0o755)
 
 
-def write_asm(asm_prelude_file: str, asm_cont: str, out_file: str) -> None:
+def write_asm(asm_prelude_file: Optional[str], asm_cont: str, out_file: str) -> None:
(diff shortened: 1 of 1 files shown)


import.py
```diff
@@ -710,8 +710,9 @@ def write_compile_command(compiler: List[str], cwd: str, out_file: str) -> None:
     os.chmod(out_file, 0o755)
 
 
-def write_asm(asm_cont: str, out_file: str) -> None:
-    with open(prelude_file, "r") as p:
+def write_asm(asm_prelude_file: str, asm_cont: str, out_file: str) -> None:
```
simonlindholm, line 713:
```suggestion
def write_asm(asm_prelude_file: Optional[str], asm_cont: str, out_file: str) -> None:
```

simonlindholm (approved):

simonlindholm:
I think we should probably have a default prelude per arch, but this seems like a reasonable thing to have regardless. (Non-MIPS support in the permuter is generally quite immature overall, so not surprised to see some issues like this.)

PR #178, opened 2024-11-06
Implement a basic speed setting to reduce the resources spent on permuting

Kelebek1:
The goal here is just that I don't want (or need) the permuter to run at max speed and taking my CPU up to max temperatures for minutes at a time, so I wanted a way to reduce the speed of execution and make it more idle friendly.

This is almost certainly not the best way to do this, so feel free to replace this with a nicer/better version, I just wanted to get *something* working.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -83,6 +83,7 @@ class Options:
     network_priority: float = 1.0
     no_context_output: bool = False
     debug_mode: bool = False
+    speed: int = 100
 
 
 def restricted_float(lo: float, hi: float) -> Callable[[str], float]:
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -738,6 +759,15 @@ def main() -> None:
         action="store_true",
         help="Debug mode, only compiles and scores the base for debugging issues",
     )
+    parser.add_argument(
+        "--speed",
+        dest="speed",
+        type=int,
+        help="Speed% to run at to reduce resources. Default 100",
+        choices=range(0,101),
+        metavar="[0-100]",
```
simonlindholm, line 768:
```suggestion
        choices=range(1,101),
        metavar="[1-100]",
```
to avoid division by zero

simonlindholm (approved):
Makes sense to me. If you're on Linux you could also use `nice` to give the permuter lower priority relative to other processes, but I don't think it helps any against overheating/fans spinning up.

PR #179, opened 2024-11-23
Attempt to fix issue WhenGryphonsFly#4 (untested)

H-A-M-G-E-R:


simonlindholm:
The error happens because you don't have a declaration for `CountPrimarySprite`. It would be nice to detect this directly and give a proper error message rather than hope that you can guess that based on the internal error! Alternatively, falling back to `int` when a function has not been declared, since it's what C does. Kinda weary of that though, feels like it just hides errors that could help match the function.

The proposed changes in this PR don't help at all, they just change the order of pattern matching cases... I would encourage testing your fixes before PRing them.

H-A-M-G-E-R:
i partially tested and i think it fixed the error

H-A-M-G-E-R:
and now i don't want my solution anymore...

simonlindholm:
Why the continuous reopen/close cycles? To be clear this PR as-is has no effect, it's effectively changing
```
if a == 1:
  do stuff 1
elif a == 2:
  do stuff 2
```
to
```
if a == 2:
  do stuff 2
elif a == 1:
  do stuff 1
```

H-A-M-G-E-R:
because i changed my mind constantly

PR #185, opened 2025-04-01
Add settings flag to specify decomp.me compiler flags

ser-pounce:
decomp-permuter can extract compiler flags that aren't necessary or even valid on decomp.me, so the flags should be overridable.

### DIFF
--- a/import.py
+++ b/import.py
@@ -728,7 +728,12 @@ def finalize_compile_command(cmdline: List[str]) -> str:
     return " ".join(quoted[:ind] + ['"$INPUT"'] + quoted[ind:] + ["-o", '"$OUTPUT"'])
 
 
-def get_compiler_flags(cmdline: List[str]) -> str:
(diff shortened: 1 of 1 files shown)


import.py
```diff
@@ -962,7 +962,7 @@ def get_setting(key: str) -> Optional[str]:
                     "context": context,
                     "source_code": source,
                     "compiler": compiler_name,
-                    "compiler_flags": get_compiler_flags(compiler),
+                    "compiler_flags": settings["decompme"]["flags"] or get_compiler_flags(compiler),
```
simonlindholm, line 965:
```suggestion
                    "compiler_flags": settings["decompme"].get("flags") or get_compiler_flags(compiler),
```
or this will crash if the property does not exist in the settings file. Or preferably, check `is not None` so that empty string does not get overridden.

Also, it would be good if you could use `json_dict()` and `assert isinstance(..., str)` so that mypy checks out.

ser-pounce: Thanks for the suggestions, I've made another attempt.

PR #188, opened 2025-06-11
Add valid User-Agent to decomp.me imports

ethteck:


ethteck:
looks like it does! Updated

simonlindholm:
Can we do lowercase `decomp-permuter`?

PR #190, opened 2025-09-02
Update labels for better support with latest splat/spimdisasm

AngheloAlf:
splat/spimdisasm added new labels on the generated assembly and changed a few defaults too (https://github.com/ethteck/splat/pull/466), breaking importing assembly to decomp-permuter.

This PR updates the labels and adds new ones to the prelude.inc file, fixing the import issue.

### DIFF
--- a/prelude.inc
+++ b/prelude.inc
@@ -1,22 +1,50 @@
 .set noat
 .set noreorder
 .set gp=64
-.macro glabel label
-    .global \label
+
+.macro glabel label, visibility=global
+    .\visibility \label
     .type \label, @function
     \label:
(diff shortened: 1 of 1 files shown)


simonlindholm:
Maybe the nonmatching macro should be defined to do nothing?

simonlindholm:
import.py does not upload object files to decomp.me, only assembly. Hm, I guess there is a point in trying to stay consistent between different macro .inc files though... How does the objdump output look with an additional .NON_MATCHING symbol, btw? I guess you've checked that permuter works as expected after this change?

PR #192, opened 2026-02-10
Add support for x86 binaries

leftmostcat:
This is mostly a cut-and-paste hackjob from asm-differ. In particular, I'm not sure I chose correctly for `branch_likely_instructions` or the return values for `process_x86_reloc`.

### DIFF
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -195,6 +195,7 @@ X86_BRANCH_INSTRUCTIONS = {
     "jpo",
     "js",
     "jz",
+    "call",
 }
 
 MIPS_SETTINGS: ArchSettings = ArchSettings(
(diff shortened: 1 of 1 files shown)


simonlindholm:
`branch_likely_instructions` should just be `set()`; it's a MIPS-specific thing. Probably good to add `call` to `branch_instructions`, but I'd leave `mov` out. `process_x86_reloc` looks fine. `--no-show-raw-insn` will break things though, it needs logic corresponding to this in asm-differ: https://github.com/simonlindholm/asm-differ/blob/4eb23bcd4bbeab81c16535c6318389e0c55584f7/diff.py#L3307-L3311

simonlindholm:
Oops, forgot about this PR!

PR #204, opened 2026-08-05
Don't raise on bare-hex relocated immediates (MIPS jump targets)

madbot-001:
`process_mips_reloc` raises on immediates that objdump prints as bare hex, which on MIPS is how jump and branch targets are printed:

```
  18:	08000047 	j	11c <func_80096928+0x11c>
			18: R_MIPS_26	.text
```

`parse_relocated_line` hands `imm = "11c"` to `process_mips_reloc`, which evaluates `int(imm, 0)` and gets `ValueError: invalid literal for int() with base 0: '11c'`.

### Why it is worth fixing rather than rare

The exception escapes `Scorer.score()`, and what happens next depends on where it lands:

* **For a candidate**, `main.py` catches it in `try_eval_candidate` and the candidate becomes an `EvalError`. The compile has already been paid for, the candidate is discarded, and with `--show-errors` off nothing is printed — so the loss is invisible.
* **For the base**, it surfaces as `CandidateConstructionFailure` and the run aborts, so any function containing such a jump cannot be permuted at all.

I hit this while A/B-testing the search loop on an N64 project (GCC 2.7.2, IDO-era code), which gave me exact numbers over 78 runs and 88,205 compiles:

* **3,927 candidates lost — 4.45% of all compiles.**
* **3 of the 16 selected functions could not be permuted**, because their base hit it.

The three bases score fine with the patch (339, 110, 251).

### The fix

`int(imm, 0)` is only used to decide whether to write a `+`, so it does not need to be strict. `imm_is_positive` tries base 0 first — decimal and `0x`-prefixed immediates keep exactly their current meaning — then base 16 for the bare-hex case, then falls back to the sign character, on the principle that a shape neither base can parse should not take the whole candidate down with it.

### Tests

Adds `test/test_objdump.py` (5 tests): the bare-hex jump target from the object above, decimal and `0x` immediates unchanged, negatives keeping their sign, zero, and an unparseable immediate not raising.

Note for reviewing: `python3 -m unittest discover -s test/` reports 16 failures both before and after this change on my machine — they are in `test_perm.py` and need a working compiler via `test/compile.sh`, which I don't have set up. The 5 new tests pass, and no previously-passing test changes state.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/objdump.py
+++ b/src/objdump.py
@@ -345,6 +345,39 @@ def pre_process(mnemonic: str, args: str, next_row: Optional[str]) -> Tuple[str,
     return mnemonic, args
 
 
+def imm_is_positive(imm: str) -> bool:
+    """Whether a relocated immediate is positive, for deciding on a '+'.
+
+    objdump prints jump and branch targets as BARE hex, e.g.
+
+        18:	08000047 	j	11c <func+0x11c>
+                18: R_MIPS_26	.text
+
+    which int(imm, 0) rejects ("invalid literal for int() with base 0:
+    '11c'"). The exception propagates out of the Scorer, main.py catches it
+    in try_eval_candidate, and the candidate is silently discarded — the
+    compile is paid for and nothing is reported. On one MIPS/IDO project it
+    cost 4.45% of all candidates (3927 of 88205 compiles), and it makes the
+    BASE unscoreable for any function containing such a jump, which is a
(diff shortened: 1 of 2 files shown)


simonlindholm:
I pushed 2795247304ec4798459b9bc865314e64e5182bf9 to the same effect.

